PR #6, opened 2018-04-21
rfc #1: store, react interop

zerobias:
Let try to discuss API before implementing them. Feel free to join the discussion if you don't understand something or if you see an issue with the approach.

### DIFF
--- a/rfc/#1. Store, react interop.md
+++ b/rfc/#1. Store, react interop.md
@@ -1,0 +1,121 @@
+# Effector RFC #1. Store, react interop
+
+- Stores for application should be as lightweight as possible.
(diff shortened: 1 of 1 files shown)


zerobias:
Streams often lead to glitches as double emits and so on.
Although they need only for debounce, we have no ability to force it yet

zerobias:
Probably `.map` is not clear enough to people who are not familiar with fp

`.transform` maybe?

zerobias:
Suggestion: add an ability to send dispatch function as the second argument (redux compatibility for example)
```createEvent('input text', store.dispatch)```

PR #39, opened 2019-02-03
Fix TypeScript bindings

miyaokamarina:
Hope adding [new type parameter](https://github.com/miyaokamarina/effector/blob/c9091c666852c7eb2dacbf61240fe5a3d9d2cd72/packages/effector/index.d.ts#L55) won’t break anything, lol.


zerobias:
Hi, everything is ok here, thanks!

Btw, don't worry, these ci errors are not related) Will fix them soon

PR #44, opened 2019-02-17
effector-react: fix a lack of rerendering between mounting and render phases

Whispers12:
Find some bug, events might have been emitted between render and mount, but effector-react doesn't handle this.

Also add test, but i don't understand why it doesn't passed


sergeysova:
@Whispers12 Thank you for contribution. Please, send link to codesandbox with reproduce

zerobias:
closed as inactive

PR #52, opened 2019-03-17
Modules

goodmind:


sergeysova:
@goodmind is it actual yet?

zerobias:
No, effector cannot benefit from es modules yet, until symbol-observable will be updated to support them as well

PR #60, opened 2019-03-20
useStore: return store state instead of react state

sergeysova:
fixes: https://codesandbox.io/s/y21jj6l8n9


doasync:
@goodmind, may @sergeysova remove useStoreFast?

doasync:
This PR should be named like 'useStore: return store state instead of react state'

doasync:
If you merge with develop there will be an unused import

zerobias:
big thanks! 👍👍

PR #62, opened 2019-03-21
docs(reamde): add a patreon icons to the header

mg901:
Add a Patreon icons to the header.


Komar0ff:
Why do we need two Patreon icons?

PR #64, opened 2019-03-22
Proposal: scopes

artalar:
`scope` is function that's return new entity with cloned dependencies. All examples and comments in the tests.


artalar:
React example:
```js
<Provider scope={createScope()}>...
// ...
const Component = () => {
  const storeScoped = useStore(useScope(store))
  const eventScoped = useScope(event)
```

zerobias:
Cloning a single entity is a wrong way to do that, as everything exists in graphs only

PR #71, opened 2019-03-30
improve store description

artalar:


README.md
```diff
@@ -242,17 +242,31 @@ The only requirement for function - **Should have zero or one argument**
 **Store** is an object that holds the state tree. There can be multiple stores.
 
 ```js
-const users = createStore([]) // <-- Default state
-  // add reducer for getUser.done event (fires when promise resolved)
-  .on(getUser.done, (state, {result: user, params}) => [...state, user])
+// `getUsers` - is a effect
```
zerobias: `a` with vowels turns into `an`: an effect, an event

README.md
```diff
@@ -242,17 +242,31 @@ The only requirement for function - **Should have zero or one argument**
...
+// `addUser` - is a event
+const defaultState = [{ name: Joe }];
+const users = createStore(defaultState)
+  // subscribe store reducers to events
+  .on(getUsers.done, (oldState, payload) => payload)
+  .on(addUser, (oldState, payload) => [...oldState, payload]))
+
+// subscribe side-effects
+const callback = (newState) => console.log(newState)
+users.watch(callback) // `.watch` for a store is triggered immidiatly: `[{ name: Joe }]`
```
zerobias: immediately

README.md
```diff
@@ -242,17 +242,31 @@ The only requirement for function - **Should have zero or one argument**
...
+// the `callback` will triggered each time when `.on` handler returns new state
```
zerobias: `callback` will be triggered each time when `.on` handler returns the new state

README.md
```diff
@@ -242,17 +242,31 @@ The only requirement for function - **Should have zero or one argument**
...
+```
+
+Most profit thing of stores is their compositions:
 
-const messages = createStore([])
-  // from WebSocket
-  .on(data, (state, message) => [...state, message])
+```js
+// `.map` accept state of parent store and return new memoized store
```
zerobias: maybe it is worth mentioning that because of this, we don't need reselect no more?

PR #85, opened 2019-04-16
Fix sampleStore, remove unnecessary watcher

doasync:
Add tests:

- store has the same state as source
- store has its own defaultState
- store updates if source updates
- store updates only if source is changed

Fix:

- Remove watch and use getState

Let's discuss sampleStore intended behaviour


doasync:
Three tests fail for current version, so we need a fix, I believe. I can remove tests from this PR, if you want.

PR #103, opened 2019-05-18
Nicer typings for createApi()

### LINKED ISSUE #102: Nicer typings for createApi()
For now createApi() doesn't let you to work with events without payload or with optional payload:

```
const api = createApi(store, {
    increment: (count) => count + 1,
    decrement: (count) => count - 1,
    double: (count) => count * 2,
    multiply: (count, mp = 2) => count * mp,

});

api.double(); // Expected 1 arguments, but got 0.
api.multiply(); // Expected 1 arguments, but got 0.
```

gzaripov:
Closes #102 
Note: `Exclude<E, undefined>` needed to remove noise from typings: `string | undefined | void` is the same as `string | void`

### DIFF
--- a/packages/effector/index.d.ts
+++ b/packages/effector/index.d.ts
@@ -251,7 +251,11 @@ export function createApi<
   store: Store<S>,
   api: Api,
 ): {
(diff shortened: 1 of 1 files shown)


zerobias:
Fix will be published in version 0.18.8

PR #116, opened 2019-06-06
Forgot import VueConstructor type

ilajosmanov:


zerobias:
Ouch :sweat_smile: Big thanks! Update will be released today

PR #122, opened 2019-07-02
sample docs added

tehSLy:
`sample` method documentation, examples added

### DIFF
--- a/docs/api/effector/sample.md
+++ b/docs/api/effector/sample.md
@@ -25,32 +25,42 @@ Returned Unit may be observed (via `watch`), since it's valid graph node.
 #### Example 1
 
 ```javascript
+import {createStore, createEvent, sample} from "effector";
+
 const store = createStore('foo');
 const event = createEvent();
 
 const sampled = sample(store, event);
-sampled.watch(console.log); // Use this watcher for side-effects
+sampled.watch(console.log);
 
 event() // => hello zerobias
 ```
-[try it](https://share.effector.dev/yQQ3y1rR)
+[try it](https://share.effector.dev/FLaJ4Nfp)
 
 #### Example 2
 
 ```javascript
-const login = createStore('peter');
+import {createStore, createEvent, sample} from "effector";
 
-const sendMessage = createEvent();
-const fullMessage = sample(login, sendMessage, (login, text) => ({login, text}));
+const login = createStore('peter')
+const sendMessage = createEvent()
 
-fullMessage.watch(({login, text}) => { // Use this watcher for side-effects
-  console.log(`[${login}]: ${text}`);
+const fullMessage = sample(
+  login,
+  sendMessage,
+  (login, text) => ({login, text}),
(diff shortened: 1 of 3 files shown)


docs/api/effector/sample.md
```diff
@@ -0,0 +1,197 @@
...
+Overall this method can be used in order to link two nodes, resulting the third one, which will fire only upon `clock` node trigger.  
+_Note: [_`Unit`_](Unit.md) chapter may come in handy, to be aware of way of using this method._
+
+Passes current `sourceStore`'s state and `clockEvent`'s value to `fn` handler. Quite a common case when you need to handle some event with some store's state. Instead of using `store.getState()`, in body of effect, which may cause race conditions and inconsistency of state at the moment of effect thunk invocation.
+
+#### Arguments
+
+1. `sourceStore` _(Store)_: Source event
+2. `clockEvent` _(Event)_: Clock(Trigger) event
+3. `fn`? _(Function)_: Callable handler (may be [_`Event`_](Event.md) or [_`Effect`_](Effect.md) aswell), should be **pure**
```
zerobias: Considering that this function exists to combine two arguments, the use of events and effects here has no meaning. Also, we can suggest to use result's own watcher for side-effects

docs/api/effector/sample.md
```diff
@@ -0,0 +1,197 @@
...
+
+#### Example
+
+```js
+const clock = createEvent();
+const inc = createEvent();
+const store = createStore(0)
+  .on(inc, (state) => state + 1);
+
+const sampled = sample(store, clock, (i) => `Current count is ${i}`);
```
zerobias: if we don't use the second argument of the handler, then there is no need for it, since the source value will be passed on by default and we can use `.map` as usual

docs/api/effector/sample.md
```diff
@@ -0,0 +1,197 @@
...
+sampled.watch(console.log);
+
+inc()
+clock(); // => Current count is 1
```
zerobias: if we show how the sample allows us to gather data across application, we'll help users to comprehend the core idea of the method

Check example in [second PR](https://github.com/zerobias/effector/pull/121/files#diff-0c4cf5585b7bed2b0d04f8b7cfb638cdR52) for sample

docs/api/effector/sample.md
```diff
@@ -0,0 +1,197 @@
...
+const greedySampling = sample({
+  source: analyticReportsEnabled,
+  clock: merge([clickButton, closeModal]),
+  fn: (isEnabled, data) => ({isEnabled, data}),
+  greedy: true,
+})
+
+commonSampling.watch(data => console.log('non greedy update', data))
+greedySampling.watch(data => console.log('greedy update', data))
+
```
zerobias: Can we provide links to an online editor for examples? Some users read the docs in markdown.

zerobias:
Great, big thanks!

PR #123, opened 2019-07-03
help Zero Bias get to console

risenforces:


zerobias:
:grin::grin:

Thanks!

PR #136, opened 2019-07-13
docs: rename thunk to handler, change createEffect examples

hexagon141:


docs/api/effector/createEffect.md
```diff
@@ -4,55 +4,116 @@ title: createEffect
...
-getUser({id: 1})
+Change state
+
+```js
+import {createStore, createEffect} from 'effector'
+
+const fetchUserRepos = createEffect({
+  handler: ({name}) => {
+    return fetch(`https://api.github.com/users/${name}/repos`)
+      .then(res => res.json())
```
zerobias: Maybe it is better to use async/await here? It could be more readable

hexagon141: > Maybe it is better to use async/await here? It could be more readable

like this?

```
const fetchUserRepos = createEffect({
  handler: async ({name}) => {
  	try {
  		const response = await fetch(`https://api.github.com/users/${name}/repos`)
  		const json = await response.json()
    	return json
  	} catch(error) {
    	return error
  	}
	}
})
```

zerobias: We don't need no try catch inside handler, that's the point of effects: we always can handle errors by `fetchUserRepos.fail`

```javascript
const fetchUserRepos = createEffect({
  async handler({name}) {
    const url = `https://api.github.com/users/${name}/repos`
    const req = await fetch(url)
    return req.json()
  }
})
```

hexagon141: > We don't need no try catch inside handler, that's the point of effects: we always can handle errors by `fetchUserRepos.fail`
> 
> ```js
> const fetchUserRepos = createEffect({
>   async handler({name}) {
>     const url = `https://api.github.com/users/${name}/repos`
>     const req = await fetch(url)
>     return req.json()
>   }
> })
> ```

Great!

zerobias:
👍 👍

PR #156, opened 2019-07-22
Update TypeScript sample typings

### LINKED ISSUE #155: Typescript typings for sample
Sample typings is incomplete for config case.

```javascript
const source = createStore<string>('')
const event = createEvent<number>()

const target = createEvent<number>()

// Fine
const case1 = sample(source, event, (a, b) => 1)

const store = createStore<number>()
const event2 = createEvent<string>()
// Should be Event<number>, not Event<string>
const case1bug = sample(event2, store, (a,b) => 1)

// No errors, but case2 should be Event, not Unit
// It doesn't matter tho, because we can use methods of target
const case2 = sample({
  source,
  clock: event,
  fn: (a, b) => 1,
  target
})
// TS complains about arguments of fn
// TS7006: Parameter 'a' implicitly has an 'any' type.
// TS7006: Parameter 'b' implicitly has an 'any' type.
const case3 = sample({
  source,
  clock: event,
  fn: (a, b) => 1
})

// And case3 is completely useless, because it has Unit type
// TS2339: Property 'filterMap' does not exist on type 'Unit<number>'.
case3.filterMap(a => a === 1 ? a : undefined)
// Can't cast it
// TS2740: Type 'Unit<number>' is missing the following properties from type 'Event<number>': watch, map, filter, filterMap, and 5 more.
const t: Event<number> = case3
```

abliarsar:
fixes #155
Probably should be a breaking change, because sample now accepts types of source and clock instead of their values

### DIFF
--- a/packages/effector/index.d.ts
+++ b/packages/effector/index.d.ts
@@ -302,43 +302,48 @@ export function restore<State extends {[key: string]: Store<any> | any}>(
 
 export function createDomain(domainName?: string): Domain
 
-export function sample<A>(config: {
-  source: Unit<A>
-  clock: Unit<any>
-  target?: Unit<A>
-}): Unit<A>
-export function sample<A, B, C>(config: {
-  source: Unit<A>
-  clock: Unit<B>
-  target?: Unit<C>
-  fn(source: A, clock: B): C
-}): Unit<C>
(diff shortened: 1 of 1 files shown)


abliarsar:
@zerobias, can you review it again? It shouldn't be breaking now.

zerobias:
Big thanks! That’s going to be a major improvement of sample user experience 👍👍

PR #162, opened 2019-07-31
Fix sandbox link in article

Laiff:
After conneting to github repo, sandbox link was changed, update it in article


Laiff:
@zerobias small fix in doc

zerobias:
👌 👌

PR #170, opened 2019-08-08
Update type of Event.filter

abliarsar:
Refine type of payload if filter got a predicate function as a callback
```javascript
import { createEvent } from 'effector'

const event = createEvent<string | number>()

const isString = (value: any): value is string => typeof value === 'string'
const isNumber = (value: any): value is number => typeof value === 'number'

const str = event.filter({fn: isString}) // Event<string>
const num = event.filter({fn: isNumber}) // Event<number>

str.watch(value => value.slice()) // OK now
num.watch(value => value.toFixed(2)) // OK now
```


zerobias:
Wow, I didn’t think it was possible 😄

PR #184, opened 2019-08-19
Examples of usage with forms in React added

tehSLy:
And usage of `getState` with replacement in favor of `sample` too!

### DIFF
--- a/docs/recipes/react/example-forms.md
+++ b/docs/recipes/react/example-forms.md
@@ -1,0 +1,192 @@
+---
+id: example-forms
+title: Example: Forms
+sidebar_label: Example: Forms
+---
+
+#### Example 1
+```js
+import React from 'react'
+import ReactDOM from 'react-dom'
+import {createEffect, createStore, createEvent, sample} from 'effector'
+import {useStore, useStoreMap} from 'effector-react'
+
+const sendForm = createEffect({handler: console.log})
+const submitted = createEvent()
+const setField = createEvent()
+const $form = createStore({}).on(setField, (s, {key, value}) => ({
+  ...s,
+  [key]: value,
+}))
+
+sample({
+  source: $form,
+  clock: submitted,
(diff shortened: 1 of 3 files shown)


docs/recipes/react/example-forms.md
```diff
@@ -0,0 +1,192 @@
...
+
+const Field = ({name, type, label}) => {
+  const value = useStoreMap({
+    store: $form,
+    keys: [name],
+    fn: values => values[name] || '',
+  })
+  return (
+    <div>
+      {label}{' '}
```
popuguytheparrot: I would use <label> instead {label}

docs/recipes/react/example-forms.md
```diff
@@ -0,0 +1,192 @@
...
+
+const sendForm = createEffect({handler: console.log})
+const submitted = createEvent()
+const setField = createEvent()
+const $form = createStore({}).on(setField, (s, {key, value}) => ({
+  ...s,
+  [key]: value,
+}))
+
+sample({
```
Tauka: How is `sample` better than
```js
const sendForm = createEffect(() =>
{
    console.log($form.getState());
}
```

and using `sendForm` directly, rather than triggering it via `submitted` event? In my example above we benefit from not creating excessive event (`submitted`)

zerobias:
Amazing!

docs/recipes/react/example-forms.md
tehSLy: `getState` is more of an escape hatch,  neither the actual way of resolving store's state. As it was mentioned in recipe's code breakdown, `getState` may cause state inconsistency and race conditions. This code still works, and even though, mostly, you will be unable to observe this kind of effect, yet in certain situations this may be the real pain in the neck.

Tauka: Thank you! Appreciate the response

PR #193, opened 2019-08-30
Add impossibilty of forwarding parts of *Effect*

Tauka:
Since *Effect* has *Events* and *Stores* within it, people may assume they can *Forward* them, as they would regular *Events* and *Stores*. However, this behavior is not implied, and hence not supported.

### DIFF
--- a/docs/api/effector/createEffect.md
+++ b/docs/api/effector/createEffect.md
@@ -18,6 +18,8 @@ Creates an [effect](Effect.md)
 
 ([_`Effect`_](Effect.md)): A container for async function.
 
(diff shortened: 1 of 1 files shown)


zerobias:
I must apologize, I had to talk about this calmer; now I'm think that forwarding really should not cause an error, although in general forwarding to parts of effect is really not supported

Tauka:
@zerobias No hard feelings :) I also apologize if I somehow disrespected your contribution, I appreciate your hard work on this library. Also, I want to make it clear that I have only positive feelings about this library and would like to further participate in its improvement. So if you have any documentation related issues or some not-so-complicated bugs, I might help out

PR #205, opened 2019-09-13
suggestion for fix #204

### LINKED ISSUE #204: Unwanted changes when using the babel plugin
Current babel plugin implementation can change `createStore`, `createEvent`, `createEffect`, `createDomain` calls even if it's wasn't imported from an effector.

Wroud:
create `Map` and store tracked imports from effector

### DIFF
--- a/src/babel/babel-plugin.js
+++ b/src/babel/babel-plugin.js
@@ -1,5 +1,12 @@
 //@noflow
 
+const definions = {
+  createStore: 'createStore',
+  createEvent: 'createEvent',
+  createEffect: 'createEffect',
+  createDomain: 'createDomain'
+}
+
 module.exports = function(babel, options = {}) {
   const {
     compressor,
@@ -20,6 +27,9 @@ module.exports = function(babel, options = {}) {
   const {types: t} = babel
   const plugin = {
     name: '@effector/babel-plugin',
+    pre() {
+      this.importedNames = new Map()
+    },
     visitor: {
       ImportDeclaration(path) {
         const source = path.node.source.value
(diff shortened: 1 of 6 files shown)


goodmind:
Can you add some tests?

Wroud:
Seems like `sid` always different. I'm tryed manually fix it because thinking that is windows trouble

zerobias:
If sid is really different on windows then it’s an important issue itself! Would you like to tell about this more? How it’s looks like?

zerobias:
Also, notice that sid will change only when you change source code; therefore it's ok that your snapshots will not match, it's assumed that you update them when you change sources.

Wroud:
It's look like this 
Screenshot
also path like this `/src/babel/__tests__/fixtures/create-component.js` on windows looks like this `\\\\\\\\src\\\\\\\\babel\\\\\\\\__tests__\\\\\\\\fixtures\\\\\\\\create-component.js`
I'm tryed change sids in snapshot from CI diff (without changing other files) but next time CI was generated new sid's

goodmind:
@Wroud can you rebase with master?

Wroud:
shure

src/babel/babel-plugin.js
```diff
@@ -64,33 +74,27 @@ module.exports = function(babel, options = {}) {
...
-            const id = findCandidateNameForExpression(path)
-            if (id) {
-              setEventNameAfter(path, state, id, babel.types, smallConfig)
-              state.domains.add(id.name)
+        if (t.isIdentifier(path.node.callee) && this.importedNames.has(path.node.callee.name)) {
+          const name = this.importedNames.get(path.node.callee.name)
+          const id = findCandidateNameForExpression(path)
+          if (id) {
+            switch (name) {
+              case definions.createStore:
```
zerobias: why do we even use it here? `definitions` are hardcoded defaults, aren't it?

src/babel/babel-plugin.js
```diff
@@ -20,6 +27,9 @@ module.exports = function(babel, options = {}) {
   const {types: t} = babel
   const plugin = {
     name: '@effector/babel-plugin',
+    pre() {
+      this.importedNames = new Map()
```
zerobias: why `this`?

Wroud: https://github.com/jamiebuilds/babel-handbook/blob/master/translations/en/plugin-handbook.md#-pre-and-post-in-plugins

src/babel/babel-plugin.js
Wroud: I'm used one common map for all tracked imports related with effector, so because we have different handlers for `createStore`, `createEvent`, `createEffect` and `createDomain`, we need match imported name with its implementation

zerobias: But CI failure implies wrong assumption here: something definitely going wrong https://github.com/zerobias/effector/pull/205/checks?check_run_id=230852210#step:5:101

Wroud: I think its nor related with my changes, i'm rebase from master now, tests snapshots failed only with `sid`s
` PASS   effector/naming  src/effector/naming/__tests__/name.test.js`

Wroud: But tests from CI was still fails hmm

Wroud: Seems babel test snapshot now match with CI, but tests wich use babel-plugin fails (but on my local machine tests pass)

zerobias:
So, eventually, you decided to silently steal the whole plugin without even mention any of the real authors from this repository and open that completely broken PR as excuse for such a rip off, didn't you?

image


Your behavior is ruining open source, go kick rocks

PR #209, opened 2019-09-25
improve typescript inference for merge method

abliarsar:
Add overload to handle cases with non-intersecting types
```javascript
const a = createEvent<number>()
const b = createEvent<string>()
const merged = merge([a,b])
```
**Current behaviour**: Error
```
Type 'Unit<string>' is not assignable to type 'Unit<number>'.
  Type 'string' is not assignable to type 'number'.
```
**After fix**: No error. `typeof merged = Event<string | number>`


zerobias:
Big thanks! 🙏 
Sorry for long delay 😅

PR #211, opened 2019-09-30
[docs] Adds case with single store argument to sample.md

Tauka:
I think it's pretty important use-case, especially if we're trying to render from multiple stores, it can save us some render cycles

### DIFF
--- a/docs/api/effector/sample.md
+++ b/docs/api/effector/sample.md
@@ -14,7 +14,7 @@ Returned Unit may be observed (via `watch`), since it's valid graph node.
 
 #### Arguments
 
-1. `sourceStore` ([_Store_](Store.md)): Source event
+1. `sourceStore` ([_Store_](Store.md)): Source store
 2. `clockEvent` ([_Event_](Event.md)): Clock(Trigger) event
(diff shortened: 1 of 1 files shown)


docs/api/effector/sample.md
```diff
@@ -256,3 +256,62 @@ clickButton('click B')
...
+import {createStore, createEffect, sample, combine} from 'effector'
+
+const data = [{name: 'physics', id: 1}]
+
+const fetchContent = createEffect({
+  handler: () => new Promise(resolve => setTimeout(() => resolve(data), 0)),
+})
+
+const $lessonIndex = createStore(0)
+const $allLessons = createStore([]).on(
```
zerobias: Tip: you can use `restore` here
https://share.effector.dev/bdAohSKM

https://github.com/zerobias/effector/blob/master/src/effector/restore.js#L34

Tauka: Sweet, didn't know about that, should I open PR for that? This one got already merged

Tauka: Or I guess, you can just push it to master, simple fix

zerobias: Just a note for the future

PR #214, opened 2019-10-06
Ts infer optional params from effect handler

bloadvenro:
The problem with current createEffect typescript typings is that [`FN extends () => infer Done`](https://github.com/zerobias/effector/blob/master/packages/effector/index.d.ts#L300) condition satisfies zero-or-one-arity effect handlers. Handler like `(x?: TYPE) => any` will produce `Effect<void,...>`. New test suite with updated snapshots [displays](https://github.com/zerobias/effector/commit/eef87cefab3068f73cc3ff2d2205484631d3bd47#diff-0015a2dd8aaa7b887423704ae136efeeR245) these errors.

You may see new typings proposal partially improving current situation. I added some comments to explain why it was difficult to deal with type `any`.

The last commit contains finally updated snapshots to demonstrate how proposed typings work.

### DIFF
--- a/packages/effector/index.d.ts
+++ b/packages/effector/index.d.ts
@@ -268,7 +268,7 @@ export function forward<To, From extends To>(opts: {
 
 export function merge<T>(events: ReadonlyArray<Unit<T>>): Event<T>
 export function merge<T extends ReadonlyArray<Unit<any>>>(
-  events: T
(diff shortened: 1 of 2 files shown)


zerobias:
Wow, amazing! 😃 Big thanks!

PR #215, opened 2019-10-07
Allow predicate refinements for split

abliarsar:
```typescript
const source = createEvent<string | number | null | {}>()

const events = split(source, {
    numbers: (a): a is number => typeof a === 'number',
    strings: (a): a is string => typeof a === 'string',
    not_refined: (a) => !!a,
})

// Event<number>, no error
events.numbers.map(a => a.toFixed())

// Event<string>, no error
events.strings.map(a => a.charCodeAt(0))

// still fails, object is possibly null
events.not_refined.map(a => a.toString())
```


zerobias:
Good improvement 👍 👍

PR #217, opened 2019-10-11
introduce `name` property in sample parameters list

bloadvenro:
Mention `name` property of `sample` parameters list right below the `guard` feature introduction.


zerobias:
Big thanks 🙏🏻

PR #219, opened 2019-10-13
docs: add documentation for guard

mg901:


zerobias:
Just a note for a future: after introduction of new chapter (`guard` in our case), we need to add it to sidebar, like [this](https://github.com/zerobias/effector/commit/f028b910d5cf02a20f102f84288075f572915dd9#diff-61f0c81048d2cc04181c023f66a432c2)

But never mind, you did great job, big thanks! 👍 👍

PR #224, opened 2019-10-16
docs: Add new example in react section

dimensi:


zerobias:
Cool! 😎

Coder2012:
Can someone explain what happens with a watch:

`submit.watch(e => e.preventDefault())`

In this example is a new watch created on every render? Do you clean up the old one in Effector for us?

zerobias:
> Can someone explain what happens with a watch:
> 
> `submit.watch(e => e.preventDefault())`
> 
> In this example is a new watch created on every render? Do you clean up the old one in Effector for us?

@Coder2012 function with this line is called outside from react components

PR #225, opened 2019-10-16
docs: Add example with dynamic form

dimensi:
Honestly, I think we should give more explanations for this example. But I don't have that opportunity at work yet, but I don't want to lose it or forget it.


zerobias:
👍 👍

PR #229, opened 2019-10-18
Better `forward` typings (TS)

bloadvenro:
Fixed wrong `forward` typings behavior, when forwarding FROM "less strict" type TO "more strict" type throws no errors.

### DIFF
--- a/packages/effector/index.d.ts
+++ b/packages/effector/index.d.ts
@@ -284,7 +284,7 @@ export function forward<T>(opts: {
   from: Unit<T & {}>;
   to: Unit<T>
 }): Subscription
-// Do not remove the second signature to avoid breaking changes!
+// Do not remove the signature below to avoid breaking change!
 export function forward<To, From extends To>(opts: {
   from: Unit<From>
   to: Unit<To>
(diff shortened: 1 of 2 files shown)


bloadvenro:
Also added tests for providing both generics (as you asked me in our community).

bloadvenro:
@abliarsar proposed an idea how to make `forward` inference even better :)

```
  it('should forward from `Unit<*>` to `Unit<void>`', () => {
    const from = createEvent<string>()
    const to = createEvent<void>()

    forward({from, to}) // <-- NOW FIXED
...
```

bloadvenro:
Remark:

```
const str = createEvent<string>()
const strOrNum = createEvent<string | number>()

// TS v3.5.3 (current version in effector repo)
//
// Produces an error:
//
// Type 'Event<string | number>' is not assignable to type 'Unit<string>'.
//     ...
forward({
  from: strOrNum,
  to: str
})

// TS v3.6.3 (current stable TS version)
//
// Produces an error:
//
// No overload matches this call.
//     ...
forward({
  from: strOrNum,
  to: str
})
```

zerobias:
Great! 👍

PR #235, opened 2019-10-29
[TS] typings fix for `combine`, accepting list of stores

bloadvenro:
Hello!

Effector 20.5.0:

```
import { combine, Store } from 'effector';

const sn = {} as Store<number>;
const ss = {} as Store<string>;

const dict = combine({ sn, ss }); // Store<{ sn: number; ss: string; }>
//                                   ^^^^^^^^^^^ That's ok! ^^^^^^^^^^^

const tuple = combine([sn, ss]); // Store<(Store<number> | Store<string>)[]>
//                                  ^^^^^^^^^^^ Looks incorrect ^^^^^^^^^^^
```

After typings fix:

```
const tuple = combine([sn, ss]); // Store<[number, string]>
//                                  ^^^^ Now that's ok too!

// As a consequence this will work after accepting the proposal
combine([sn, ss], ([n, s]) => {
  n.toExponential; // OK, has type specific method
  s.charAt; // OK, has type specific method
});
```

### DIFF
--- a/packages/effector/index.d.ts
+++ b/packages/effector/index.d.ts
@@ -1,5 +1,13 @@
 /// <reference types="symbol-observable" />
 
+/**
+ * This tuple type is intended for use as a generic constraint to infer concrete
+ * tuple type of ANY length.
+ *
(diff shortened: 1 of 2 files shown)


zerobias:
Thanks! Feel free to open new pr in case of additional edge cases

PR #236, opened 2019-10-30
added custom wrapper for vue package. access to store from this

ilajosmanov:
```
const counter = createStore(0);
export default createComponent({
  mounted() {
    console.log(this.counter); // number
  }
}, { counter });
```

With this wrapper, we have access to a store with TS typing.


ilajosmanov:
Do you mean just build package with yarn?

zerobias:
Yep, just remove package-lock.json and it will be ok

zerobias:
Great, thanks! 👍

PR #241, opened 2019-11-02
Create typings test suite for `restore`

### LINKED ISSUE #239: [TS] restore<T | null>(Effect<any, T>, null) fails due to contravariance
Hello!

Effector v20.5.0
TS v3.6.3

Consider the case:

```
declare const fx: Effect<any, number>;

restore(fx, 0); // OK [correct]
restore<number>(fx, 0); // OK [correct]

restore<number | null>(fx, null); // Error, but should be OK.
```

TS gives the following error:

```
Argument of type 'Effect<any, number, Error>' is not assignable to parameter of type 'Effect<any, number | null, any>'.
  Types of property 'done' are incompatible.
    Type 'Event<{ params: any; result: number; }>' is not assignable to type 'Event<{ params: any; result: number | null; }>'.
      Types of parameters 'payload' and 'payload' are incompatible.
        Type '{ params: any; result: number | null; }' is not assignable to type '{ params: any; result: number; }'.ts(2345)
```

As you can see, TS digs into `Effect<...>` type internals and throws the error on `Effect<...>.done` assignment.

This happens due to `strict: true`, which includes `strictFunctionTypes: true`. If I set `strictFunctionTypes: false`, the above error will dissapear. In other words "contravariant function types" behavior does not allow assign a more strict `payload` type to a less strict one.

bloadvenro:
Hello! It seems like there is no test suite for `restore` typings. I've created one. Could you please review?

If you accept this, I'll start working on #239 . The second `describe` section in the file has corresponding failing tests. The thitd `describe` section has tests which are nessesary to make sure that future fix won't break basic `restore` behavior.

### DIFF
--- a/src/types/__tests__/effector/restore.test.js
+++ b/src/types/__tests__/effector/restore.test.js
@@ -1,0 +1,208 @@
+// @flow
+/* eslint-disable no-unused-vars */
+import {restore, createEvent, createEffect, createStore, Store} from 'effector'
+import setupLocation from '../../setupLocation'
+const typecheck = '{global}'
+
+describe('restore cases (should pass)', () => {
+  test('restore(Event<number>, number): Store<number>', () => {
+    const ev = createEvent<number>()
+
+    let restored = restore(ev, 0)
+    let store: Store<number>
+
+    store = restored
+    restored = store
+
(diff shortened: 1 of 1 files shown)


zerobias:
Great! Sure, you can start working on that issue 👍

PR #245, opened 2019-11-06
Docs: update combine docs

dimensi:
With `?`, I'm trying to show you that the function is not necessary.

### DIFF
--- a/docs/api/effector/combine.md
+++ b/docs/api/effector/combine.md
@@ -4,7 +4,7 @@ title: combine
 hide_title: true
 ---
 
-# `combine(...stores, fn?)`
+# `combine(...stores, fn)`
 
 Creates a new [store](Store.md) that emits the set of latest store values from all input stores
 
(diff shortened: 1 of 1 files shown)


docs/api/effector/combine.md
```diff
@@ -4,7 +4,7 @@ title: combine
 hide_title: true
 ---
 
-# `combine(...stores, fn)`
+# `combine(...stores, fn?)`
```
zerobias: It’s not correct: you can use either object/array with stores (one argument, and optional function) or multiple arguments in which each argument is store and the last is a function (not optional this time)
we have each real case covered https://github.com/zerobias/effector/blob/master/src/effector/__tests__/combine.test.js#L15

in short: this overload should not change, but all other changes are ok 👍

dimensi: Actually, this signature confuses me with `combine(...store, fn)`. I thought that fn was necessary, but as it turned out, if you cast it, you get an array.

zerobias:
That’s good, thanks! 👍

PR #248, opened 2019-11-15
Sample target forwarding typings

### LINKED ISSUE #247: [TS] `sample` with `target`: "best common type" problem
```ts
sample({
  target: {} as Unit<{ a: string; b: string }>,
  source: {} as Unit<{ a: string }>, // should not let such assignment
  //////////////////////////////////////////////////////////////////////////////
  ///////////////// types of props below do not affect anything ////////////////     
  //////////////////////////////////////////////////////////////////////////////
  clock: {} as Unit<any>,
});

sample({
  target: {} as Unit<{ a: string; b: string }>,
  fn: () => ({ a: '' }), // should not let such assignment
  //////////////////////////////////////////////////////////////////////////////
  ///////////////// types of props below do not affect anything ////////////////     
  //////////////////////////////////////////////////////////////////////////////
  source: {} as Unit<any>,
  clock: {} as Unit<any>,
});
```

Last time we faced with the same problem in `forward` typings.

I'll add `NoInfer<T>` type [suggested](https://github.com/microsoft/TypeScript/issues/14829#issuecomment-504042546) by Typescript comminuty and modify inference for each `sample` overload which includes `target`.

bloadvenro:
+ test suit for `sample` cases using `target` forwarding
+ add NoInfer type hint to `sample` overloads using `target`

This will fix #247

### DIFF
--- a/packages/effector/index.d.ts
+++ b/packages/effector/index.d.ts
@@ -8,6 +8,13 @@
  */
 type Tuple<T = unknown> = [T] | T[]
 
+/**
+ * Non inferential type parameter usage.
+ *
+ * @see https://github.com/microsoft/TypeScript/issues/14829#issuecomment-504042546
+ */
+type NoInfer<T> = [T][T extends any ? 0 : never]
+
 export const version: string
 
 export type kind = 'store' | 'event' | 'effect' | 'domain'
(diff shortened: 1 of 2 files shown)


zerobias:
Cool, thanks! 👍

PR #273, opened 2019-12-18
Improve doc visibility for mobile

YanLobat:
Updated top-level exports with couple features from the sidebar


YanLobat:
Will that work out? @zerobias

zerobias:
Yeah, sure!

PR #274, opened 2019-12-18
Improve useStoreMap typings

abliarsar:
This uses tuple hack we already used for `combine` to improve inference for array of keys


zerobias:
Wow! 👍 👍

PR #276, opened 2019-12-21
[Typings] Workaround for typescript issue 35162

bloadvenro:
Workaround for https://github.com/microsoft/TypeScript/issues/35162 (tharks to @artalar for help)

Now we can directly assign `guard` invocation to `sample`'s `clock` argument without losing inference in `sample`'s `fn`

```ts
import {sample, guard, createStore, createEvent} from 'effector'

const source = createStore(100)
const clock = createEvent<number>()

sample({
  source,
  clock: guard(clock, {
    filter: clock => clock > 0,
  }),
  fn: (source, clock) => source + clock, // clock is of type `number` now (previously inferred as `unknown`)
})
```

### DIFF
--- a/packages/effector/index.d.ts
+++ b/packages/effector/index.d.ts
@@ -90,6 +90,16 @@ export interface Event<Payload> extends Unit<Payload> {
   [Symbol.observable](): Observable<Payload>
 }
 
+/**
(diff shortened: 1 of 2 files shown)


zerobias:
Amazing!

PR #281, opened 2019-12-24
Change theme. Increase code-font-size

popuguytheparrot:
Change prism theme to vsDark
Increase css var  `code-font-size` to 95%


zerobias:
👍

PR #284, opened 2019-12-31
Sample multitarget TS typings [WIP]

bloadvenro:
Add sample multitarget typings support [WIP]

### DIFF
--- a/packages/effector/index.d.ts
+++ b/packages/effector/index.d.ts
@@ -20,6 +20,45 @@ type Combinable = { [key: string]: Store<any> } | Tuple<Store<any>>
 // Helper type, which unwraps combinable sample source value.
 type GetCombinedValue<T> = {[K in keyof T]: T[K] extends Store<infer U> ? U : never}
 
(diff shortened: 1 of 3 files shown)


zerobias:
Released in [`effector 21.8.0`](https://github.com/effector/effector/releases/tag/effector%4021.8.0)

I was looked at this PR during whole year before realized what exactly goes wrong: we don't need to describe whole amount of cases, we just need to generate them, so I created type test generator based on your cases. It accepts [config](https://github.com/effector/effector/commit/1ad02f54e546662f24339c439c7b877b3bfef595#diff-adc625a87bf9fbad71288d3fdd991f804866e8c70b590858515213c6b1b830c1R35) and yields [a lot of permutations](https://github.com/effector/effector/commit/1ad02f54e546662f24339c439c7b877b3bfef595#diff-d26be224d21d320027ba78e9095423427b87c7bc97add2a33e988e0a020bc81eR169) based on it. Without manual work

Big thanks for inspiration and really impressive work! 👍

PR #300, opened 2020-02-06
added [stop] and [prevent] property for configure listener from [handler]

ilajosmanov:
When we configure a handler for a tag we can set preventDefault and stopPropagation functions.
```js
const click = createEvent()

/* Before: */
click.watch(e => e.preventDefault())
h('button', () => {
  handler({passive: false}, {click})
  spec({text: 'Click me'})
})

/* After: */
h('button', () => {
  handler({prevent: true /* also you could set `stop: true` from stopPropagation  */}, {click})
  spec({text: 'Click me'})
})
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,15618 @@
+{
+  "name": "effector-dev",
+  "version": "0.0.1",
+  "lockfileVersion": 1,
+  "requires": true,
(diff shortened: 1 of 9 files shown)


zerobias:
Great, useful feature 👍

PR #301, opened 2020-02-09
created option [fn] like cb

ilajosmanov:
Some else way to render items from `list` function. effector like the way (sample, guard)

```js
const $users = createStore([
  {name: 'Alice', id: 1},
  {name: 'Bob', id: 2}
])

list({
  source: $users,
  key: 'id',
  fn: ({store}) => {
    h('span', {
        text: store.map(({name}) => name)
    })
  }
})
```


zerobias:
Amazing, thanks! 👍

PR #302, opened 2020-02-11
WIP: Fix argument name in split method

Sozonov:


zerobias:
I'm think that change is good as it is, so I'm going to merge it, thanks! 👍 We always can clarify definition even more in separate PR

PR #303, opened 2020-02-11
docs page : update createEffect names according naming convention, fix incorrect imports in Store examples

egaris:


zerobias:
Fantastic! Documentation becomes much more consistent 👍👍

PR #306, opened 2020-02-15
fix(effector): replace two handlers of inFlight with one

sergeysova:


yumauri:
This could be breaking change, because this changes order of events, from

> `finally` → `done`/`fail` → `inFlight`

to

> `finally` → `inFlight` → `done`/`fail`

Looks like effector doesn't have tests for that case, but my tests in ReEffect will definitely need reconcideration :)

And to be clear:
I'm OK with this PR, just making point on changes.

yumauri:
I thought a bit more, maybe I am wrong about order... `done`/`fail` remains first and second children of `finally`, and `inFlight` is added as third one, and this will not change order. Need to check explicitly.

zerobias:
@yumauri 

> This could be breaking change, because this changes order of events, from
> 
> > `finally` → `done`/`fail` → `inFlight`
> 
> to
> 
> > `finally` → `inFlight` → `done`/`fail`

I added inFlight and pending to the [current execution order test](https://github.com/zerobias/effector/blob/master/src/effector/__tests__/effect/index.test.js#L271). Indeed, this PR will change execution order: `doneData` will be called after `inFlight 0`. Good that we didn't release `doneData` yet 🙃

zerobias:
Nice catch, @sergeysova! I was overlooking that move

PR #309, opened 2020-02-19
[Docusaurus] improvs and fixs

popuguytheparrot:
- up version
- add light theme for code
- back example on main page
- change sh to bash in docs for right display of highlights


zerobias:
Great, thanks! 👍 👍

PR #312, opened 2020-02-23
Small Effect changes

yumauri:
Main thing I want to change here → preserve _any_ fields in result in events `done` and `fail`.

I don't know if you are ok with destructuring assignment syntax, if no — feel free to reject or ask for changes :)

### DIFF
--- a/src/effector/createEffect.js
+++ b/src/effector/createEffect.js
@@ -33,33 +33,31 @@ export function createEffect<Payload, Done>(
     handler = fn
     return instance
   }
-  const getHandler = instance.use.getCurrent = () => handler
-  const anyway = instance.finally = createNamedEvent('finally')
+  const getHandler = (instance.use.getCurrent = () => handler)
+  const anyway = (instance.finally = createNamedEvent('finally'))
 
-  const omitStatus = (data) => {
-    const result = { ...data }
+  const omitStatus = data => {
+    const result = {...data}
     delete result.status
     return result
   }
-  const done = instance.done = filterMapEvent(anyway, {
+  const done = (instance.done = filterMapEvent(anyway, {
     named: 'done',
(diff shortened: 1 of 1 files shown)


yumauri:
Hm. I see some build issues...

yumauri:
Changed to good old `delete`.
I don't really like `omitStatus` function, but what do you think about whole approach?

yumauri:
I wanted to reuse these events in ReEffect :)
But I cannot in the current implementation, because data can contain field `strategy`, which is eliminated in `done`/`fail` after mapping :(

yumauri:
I understand this is bold change of internal implementation in favor of external tool, so, I'm totally fine if you reject this approach. But it was worth trying :)

yumauri:
Oh, thank you, I didn't pay attention, that stack contains previous values.

Returned mapping for `done`/`fail` event, what about other changes?

I removed bind for `onSettled`, because regular functions are more performant, that binded ones.

yumauri:
Also I'm curious, why rename `finally` twice? Just to have name "finally" inside `onCopy`, corresponding with "finally" name of event?

image

zerobias:
Could you show a difference in performance between binded and curried functions?

zerobias:
Yes, finally in scope must match with name in onCopy array

yumauri:
Unfortunately I'm not good at micro-benchmarking, but I was sure this is common knowledge. Maybe it is not so clear in the modern engines, but binded functions was definitely slower maybe few years ago. 

There are two benchmarks (not mine) for example, and results in my Safari right now:

https://jsperf.com/bind-vs-closure-2019

image

https://jsperf.com/partial-bind-closure

image

zerobias:
Seems reasonable, thanks!

PR #313, opened 2020-02-29
fix effector-vue. Now work with component watch opts

Fl0pZz:


zerobias:
Splendid! Big thanks! 👍👍

PR #318, opened 2020-03-21
Update code example

yumauri:
Slightly updated code example, in the old version it was not obvious that reducer functions actually works:

```javascript
add(2)
// → counter: 2
sub(1)
// → counter: 1
```

looks like event just passes value to the store.
I changed so it should be more clear:

```javascript
add(5)
// → counter: 5
sub(1)
// → counter: 4 ← now we see that reducer actually works
```


zerobias:
Thanks!

PR #319, opened 2020-03-22
[Proof-of-Concept] Get rid of bind* functions

yumauri:
This PR is just to show the differences, you can take an idea :)


src/effector/createUnit.js
```diff
@@ -279,7 +278,7 @@ export function createStore<State>(
       } else {
         eventOrFn(store.getState())
       }
-      return watchUnit(store, eventOrFn)
+      return watchUnit(store)(eventOrFn)
```
zerobias: No `fn()()`, please

yumauri: What about like this? :)
```suggestion
      const watchStore = watchUnit(store)
      return watchStore(eventOrFn)
```

zerobias: Nice try 😄But no, no `fn()()`, please

yumauri: Hey, you have `fn()()` inside `createEffect` ;) And that wasn't my work :)

image

yumauri:
Reckon, idea was taken and considered, so I'll close this PR to clear some space and reduce amount of chaos in the universe :)

PR #323, opened 2020-03-25
add type guard for target in guard

sergeysova:


src/effector/guard.js
```diff
@@ -25,6 +25,9 @@ export function guard(source, config) {
   const {filter, greedy, name = 'guard'} = config
   const target = config.target || createEvent(name)
   if (!is.unit(source)) source = combine(source)
+  if (!is.unit(target)) {
```
zerobias: Target can be an array https://github.com/zerobias/effector/pull/284

PR #327, opened 2020-04-22
[Docs] Reviewing english structure and grammar

Tauka:
This PR fixes language problems, both structural and grammatical. It did not fix some other problems, such as using old api in some examples, they should be addressed in separate PR


zerobias:
Amazing! A huge improvement for our docs 👍👍👍

PR #328, opened 2020-04-23
Add support for trigger[] in store.on

sergeysova:
```js
import {createEvent, createStore} from 'effector'

const store = createStore(0)
const changedA = createEvent()
const changedB = createEvent()

store.on([changedA, changedB], (state, params) => state + params)

store.watch(value => {
  console.log('updated', value)
})

changedA(2)
// => updated 2

changedB(2)
// => updated 4

store.off(changedB)
```

### DIFF
--- a/src/effector/createUnit.js
+++ b/src/effector/createUnit.js
@@ -17,6 +17,7 @@ import {launch} from './kernel'
 import type {Subscriber, Config} from './index.h'
 import {createName, mapName, joinName} from './naming'
 import {createLinkNode} from './forward'
+import {merge} from './merge'
 import {watchUnit} from './watch'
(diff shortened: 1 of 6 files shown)


src/effector/createUnit.js
```diff
@@ -187,12 +187,14 @@ export function createStore<State>(
...
-      store.off(event)
-      getSubscribers(store).set(
-        event,
-        createSubscription(updateStore(event, store, 'on', true, fn)),
-      )
+    on(events, fn) {
+      if (!Array.isArray(events)) {
+        onEvent(events, fn)
+      } else {
+        for (const event of events) {
```
zerobias, line 194: Nice catch! I didn’t even think about it that way

PR #329, opened 2020-04-24
remove inequality operator

hexagon141:
Maybe inequality operator not needed there? 🧐


zerobias:
Alright, seems reasonable, thanks! 👌

PR #348, opened 2020-06-02
add sample overloads; add test type check;

egorguscha:


src/types/__tests__/effector/sample.test.js
```diff
@@ -946,3 +946,40 @@ describe('without clock', () => {
...
+    sample({
+      source,
+      clock,
+      target,
+    })
+
+    expect(typecheck).toMatchInlineSnapshot(`
+      "
+      --typescript--
+      No overload matches this call.
```
zerobias, line 965: Working overload means that this test should pass, please improve this.

Note that `A extends Combinable` means that such overload is for `source: {foo, bar}`, what imply irrelevant test: it not checking what is really changed

PR #350, opened 2020-06-07
Fine-grained effector-vue watch binding

ilajosmanov:
Access to pass an event to `effector` object. This event will return timestamp when called this event.

Motivation:

Handle when called this event and prevent this call on `watch` (vue API)

For example: 
I want to send data and then close popup on the view layer.

```
const send = createEffect({...});

export default {
  effector: {
    completed: send.done
  },
  watch: {
    completed() { this.closePopup() }
  }
}
```

### DIFF
--- a/src/vue/__tests__/index.test.ts
+++ b/src/vue/__tests__/index.test.ts
@@ -1,5 +1,5 @@
 import {shallowMount, createLocalVue} from '@vue/test-utils'
-import {createStore} from 'effector'
+import {createStore, createEvent} from 'effector'
 import Vue from 'vue'
 import {VueEffector, createComponent} from 'effector-vue'
 
@@ -67,6 +67,23 @@ test('vue $watch', async() => {
 
   expect(wrapper.vm.counter).toBe(1)
   expect(mockCallback.mock.calls.length).toBe(1)
+})
+
+test('test SID if pass event to effector object', async() => {
+  const evt = createEvent()
+
+  const component = createComponent({
+    template: '<div>{{ evt }}</div>',
+  }, {evt})
+
+  const wrapper = shallowMount(component, {
+    localVue
+  })
+
+  evt()
+  // @ts-ignore
+  console.log(wrapper.vm.evt)
+  expect(wrapper.vm.evt).not.toBeNull()
 })
 
 test('vue component watch option', async() => {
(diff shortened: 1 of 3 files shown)


src/vue/index.ts
```diff
@@ -78,8 +78,20 @@ const effectorMixin: {
...
       } else if (typeof shape === 'object' && shape !== null) {
-        const store = combine(shape)
-        for (const key in shape) {
+        const state = Object.keys(shape)
+          .reduce((acc, key) => {
+            const k = key as keyof typeof shape;
+            const v = shape![k];
+
+            return {
+              ...acc,
```
zerobias, line 87: No object spread in cycle, please

```typescript
const state = {} as Record<string, Store<any>>
for (const key in shape) {
  const v = shape[key]
  state[key] = is.store(v) ? v : restore(v.map(() => Date.now()), null)
}
```

src/vue/__tests__/index.test.ts
```diff
@@ -69,6 +69,30 @@ test('vue $watch', async() => {
...
+
+
+  const component = createComponent({
+    template: '<div>{{ evt }}</div>',
+    watch: {
+      evt() {
+        mockCallback()
+      }
+    }
+  }, {evt})
```
zerobias, line 84: Maybe it's a time to merge both arguments into one config? Like we did in `forest`

```typescript
const component = createComponent({
  template: '<div>{{ evt }}</div>',
  event: {evt},
  watch: {
    evt() {
      mockCallback()
    }
  },
})
```

src/vue/index.ts
```diff
@@ -78,8 +78,20 @@ const effectorMixin: {
...
+              // @ts-ignore
+              [key]: is.store(v) ? v : restore(v.map(() => Date.now()), null)
```
zerobias, line 89:
```typescript
let nextID = 0 // module-level variable

...

v.map(() => ++nextID)
```

ilajosmanov: Make scence!

src/vue/__tests__/index.test.ts
ilajosmanov: no-no. vue has the same section. it's `methods`.
in this case, we testing, what happened when we pass an event to store object

src/vue/index.ts
ilajosmanov: got it

PR #356, opened 2020-06-12
Fix in sample api

hyposlasher:
If the `clock` is not passed, sample will be trigged on every `source` update (instead of `target`)


zerobias:
Nice catch 👍

PR #358, opened 2020-06-18
console.log in REPL does not render Error objects in a right way

oas89:
**Current behavior**: `Error` objects are rendered in REPL as empty `{}` object.
**Reproduction**: https://share.effector.dev/T0okfYak


zerobias:
Cool, thanks! 👍 👍

PR #360, opened 2020-06-25
small accessability fix for the website

leushkin:
Those images don't need to have an alt tag, because they are decorative images. And by the way, it's also wrong to set an alt with text which is already on the screen cuz screen reader users will hear this text twice and that makes no sense.


zerobias:
Seems reasonable, thanks 👍

PR #364, opened 2020-07-04
Lexend font. Update docusaurus

popuguytheparrot:


zerobias:
Cool, I like new font 👍👍

PR #366, opened 2020-07-04
Update attach.md

dimensi:
Little error in code example


zerobias:
I think this was supposed to have `return req.json()` at the end. But anyway, that change seems reasonable too, nice catch, thanks 👍

PR #376, opened 2020-08-13
add filter and filterMap types to Effect

sergeysova:
```js
const fx = createEffect().use(() => null)

const d = fx.filter({
  fn(a) {
    return a > 0
  },
})

d.watch(console.log)

fx(5)
fx(0)
fx(1)
```

https://share.effector.dev/G34PZ7qF

### DIFF
--- a/packages/effector/index.d.ts
+++ b/packages/effector/index.d.ts
@@ -139,6 +139,8 @@ export interface Effect<Params, Done, Fail = Error> extends Unit<Params> {
   pending: Store<boolean>
   inFlight: Store<number>
   watch(watcher: (payload: Params) => any): Subscription
+  filter(config: {fn(payload: Params): boolean}): Event<Params>
+  filterMap<T>(fn: (payload: Params) => T | undefined): Event<T>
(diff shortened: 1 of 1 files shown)


packages/effector/index.d.ts
```diff
@@ -139,6 +139,8 @@ export interface Effect<Params, Done, Fail = Error> extends Unit<Params> {
   pending: Store<boolean>
   inFlight: Store<number>
   watch(watcher: (payload: Params) => any): Subscription
+  filter(config: {fn(payload: Params): boolean}): Event<Params>
+  filterMap<T>(fn: (payload: Params) => T | undefined): Event<T>
```
doasync, line 143: image

doasync: Is there a reason why you have skipped the first overload?

zerobias:
👍

packages/effector/index.d.ts
sergeysova: It is forgotten.

PR #380, opened 2020-08-26
docs(website): fix effect fn signature in example

jsejcksn:


zerobias:
Nice catch, thanks! 👍

PR #392, opened 2020-10-02
Refactor examples in docs, add actual links to repl

oas89:


zerobias:
Examples becomes much better, thanks 👍

PR #400, opened 2020-10-09
[TS] Void function interop

doasync:
Use case:
```ts
  type voidFn = () => void
  const fn = (callback: voidFn) => {
    if(callback) callback()
  }
  const voidEvent: Event<void> = createEvent();
  fn(voidEvent); // TS2345: Argument of type 'Event ' is not assignable to parameter of type 'voidFn'.
```

This PR adds void function interop:

```ts
  const voidFn: () => void = createEvent<void>()
  const voidFn1: () => void = createEvent<undefined>()
  const voidFn2: () => void = createEvent<any>()

  const event1 = createEvent<undefined>()
  event1()
  const event2 = createEvent<any>()
  event2()
```

The only weak spot is the text of the message:

```ts
  const event1 = createEvent<number>()
  event1() /* Error
    The 'this' context of type 'void' is not assignable to method's 'this' of type '"Error:
	Expected 1 argument, but got 0"'.
  */
```

All tests pass!

### DIFF
--- a/packages/effector/index.d.ts
+++ b/packages/effector/index.d.ts
@@ -83,6 +83,7 @@ export type CompositeName = {
 
 export interface Event<Payload> extends Unit<Payload> {
   (payload: Payload): Payload
+  (this: Payload extends void ? void : `Error: Expected 1 argument, but got 0`): void
   watch(watcher: (payload: Payload) => any): Subscription
   map<T>(fn: (payload: Payload) => T): Event<T>
   filter<T extends Payload>(config: {
--- a/src/types/__tests__/effector/effect.test.js
+++ b/src/types/__tests__/effector/effect.test.js
@@ -355,9 +355,6 @@ describe('#filterMap', () => {
       "
       --typescript--
       Type 'Event<number>' is not assignable to type 'Event<number | void>'.
-        Types of parameters 'payload' and 'payload' are incompatible.
-          Type 'number | void' is not assignable to type 'number'.
-            Type 'void' is not assignable to type 'number'.
 
       --flow--
       no errors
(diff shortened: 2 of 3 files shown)


packages/effector/index.d.ts
```diff
@@ -83,6 +83,7 @@ export type CompositeName = {
 
 export interface Event<Payload> extends Unit<Payload> {
   (payload: Payload): Payload
+  (this: Payload extends void ? void : `Error: Expected 1 argument, but got 0`): void
```
zerobias, line 86: What is this?

zerobias (commented):
Tests indeed failed

packages/effector/index.d.ts
doasync: "this" is a type of context, it is void by default
CleanShot 2020-10-09 at 08 15 57@2x

zerobias: What is \`Error: Expected 1 argument, but got 0\`?

doasync: It's just a text to show in the error message

doasync: According to tests, it will be shown only if Event expects an argument, but no parameters given.

src/types/__tests__/effector/effect.test.js
```diff
@@ -355,9 +355,6 @@ describe('#filterMap', () => {
       "
       --typescript--
       Type 'Event<number>' is not assignable to type 'Event<number | void>'.
```
doasync, line 357:
```js
Type 'Event<number>' is not assignable to type 'Event<number | void>'
```

It is important to put "this" overload after the first one to keep this message right.

doasync:
I've updated the snapshot

zerobias:
Interesting solution, okay, lets try it. Thanks! 👍

PR #406, opened 2020-10-19
Fix guard return type (support union type)

doasync:


zerobias:
Cool improvement, thanks 👍  I'm wonder if it would help in other similar cases?

PR #408, opened 2020-10-21
Add warning about hydration requiring initialized stores

bonkboykz:
Update hydrate.md


zerobias:
Looks reasonable, thanks 👌 Although this is a common property of whole fork api

PR #409, opened 2020-10-21
add patronum to repl

sergeysova:


website/editor/src/evaluator/index.tsx
```diff
@@ -137,6 +137,17 @@ const fetchEffectorReactSSR = createEffect({
...
 })
 
+const fetchPatronum = createEffect({
+  async handler(effector) {
+    const url = 'https://unpkg.com/patronum/patronum.cjs.js'
+    const sourceMap = `${url}.map`
+    const req = await fetch(url)
+    let text = await req.text()
+    text = text.replace(/\/\/\# sourceMappingURL\=.*$/m, `//${tag}MappingURL=${sourceMap}`)
+    return createRealm(text, `ssr.js`, {effector})
```
zerobias, line 147: `'patronum.js'`

This became a name for that module in browser devtools

website/editor/src/evaluator/index.tsx
```diff
@@ -276,8 +290,15 @@ const removeImportsPlugin = babel => ({
         case 'effector-react/ssr':
           replaceModuleImports('effectorReactSSR', path, babel)
           break
+        case 'patronum':
+          replaceModuleImports('patronum', path, babel)
+          break;
         default:
-          path.remove()
+          // TODO: import actual module after patronum refactoring to ES Modules
```
zerobias, line 297: I think we can avoid additional modules altogether, reusing them like that is indeed a good thing

zerobias (changes requested):
Please update module name from `ssr.js` to `patronum.js`

website/editor/src/evaluator/index.tsx
sergeysova, line 147:
```suggestion
    return createRealm(text, `patronum.js`, {effector})
```

zerobias:
Big thanks! Now we can creat much more advanced and neat examples for documentation 😃

zerobias (approved):

PR #414, opened 2020-11-04
Fix event type, add type tests

doasync:


src/types/__tests__/effector/effect.test.js
```diff
@@ -487,6 +487,9 @@ describe('#filterMap', () => {
       "
       --typescript--
       Type 'Event<number>' is not assignable to type 'Event<number | void>'.
+        Types of parameters 'payload' and 'payload' are incompatible.
```
zerobias, line 490: Interesting, changes in this pr makes that case more precise 👍

zerobias:
Good improvement, thanks!

PR #415, opened 2020-11-05
Add type tests for prepend, improve prepend type

doasync:
`prepend` method infers payload type from the function argument if it exists... and if there is no argument the resulting payload type will be of type `unknown`. This is a bug because a user cannot call `Event<unknown>` without params -  it waits for 1 argument with type `unknown`.

```ts
const event = createEvent<string>();
const prepended = event.prepend(() => 'foo') // Event<unknown>
prepended() // TS error: Expected 1 arguments, but got 0.
```

To bypass the type error, a user needs to specify `void` manually: 
```ts
const event = createEvent<string>();
const prepended = event.prepend<void>(() => 'foo') // Event<void>
prepended() // ok
```

But we can fix this by adding a default type to prepend's `Before`:
```ts
  prepend<Before = void>(fn: (_: Before) => Payload): Event<Before>
```

As the result, TypeScript will show an error when you directly call `Event<void>` with an argument, which is good too:

```ts
const event = createEvent<string>()
const prepended = event.prepend(() => 'foo')
prepended(1) // Argument of type 'number' is not assignable to parameter of type 'void'.
```

Now it can be seen that the addition of the default value improves typing in both cases.

P.S. I also added type tests for `.prepend` and the described cases.

### DIFF
--- a/packages/effector/index.d.ts
+++ b/packages/effector/index.d.ts
@@ -116,7 +116,7 @@ export interface Event<Payload> extends Unit<Payload> {
   }): Event<T>
   filter(config: {fn(payload: Payload): boolean}): Event<Payload>
   filterMap<T>(fn: (payload: Payload) => T | undefined): Event<T>
-  prepend<Before>(fn: (_: Before) => Payload): Event<Before>
+  prepend<Before = void>(fn: (_: Before) => Payload): Event<Before>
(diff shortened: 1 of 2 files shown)


zerobias:
Cool idea! 🚀

PR #419, opened 2020-11-08
[Vue] marked deprecated methods

ilajosmanov:


zerobias:
Okay, but now we need to add warning to code as well. Anyway, thanks 👍

PR #420, opened 2020-11-10
Deprecate $watchAsStore and $store

Drevoed:


zerobias (changes requested):
Please add also console notification on methods' calls, like this one:

```javascript
console.error('$watchAsStore is deprecated')
```

PR #426, opened 2020-11-22
allow to use `remap` without `as const`

sergeysova:
```tsx
const $data = createStore({
  foo: 1,
  bar: "two",
  baz: true,
})

const d: [
  Store<number>,
  Store<boolean>,
  Store<string>,
] = remap($data, ['foo', 'baz', 'bar'])
```


zerobias:
Wow, I didn't even think it was possible 👍👍

PR #431, opened 2020-12-09
Fix/useStore

zarabotaet:
Two different workarounds in two commits

### DIFF
--- a/src/react/__tests__/useStore.test.tsx
+++ b/src/react/__tests__/useStore.test.tsx
@@ -90,7 +90,7 @@ describe('useStore', () => {
       ]
     `)
   })
-  it('should always retrigger, when store contains function', async () => {
+  it('should correct work, when store contains function', async () => {
     const fn = jest.fn()
     const changeStore = createEvent()
     const $store = createStore(() => 0).on(changeStore, (_, p) => p)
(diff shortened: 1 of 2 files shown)


zarabotaet:
Because useState is a wrapper  around useReducer we can directly take useReducer.
https://codesandbox.io/s/hopeful-lovelace-qeufx?file=/src/App.js
Sandbox to play with the proposed solution.

src/react/__tests__/useStore.test.tsx
```diff
@@ -90,6 +90,29 @@ describe('useStore', () => {
...
     `)
   })
+  it('should correct work, when store contains function', async () => {
+    const fn = jest.fn()
+    const changeStore = createEvent()
+    const $store = createStore(() => 0).on(changeStore, (_, p) => p)
+
+    const Display = () => {
+      const store = useStore($store)
+      fn()
```
zerobias, line 100: There should be a proof that issues does exists at all: your example [reveal](https://codesandbox.io/s/admiring-goldberg-i9q5l?file=/src/App.js) that current implementation works as expected

zerobias (changes requested):

src/react/__tests__/useStore.test.tsx
zarabotaet: [original issue](https://share.effector.dev/5T7AGosk) 
[new sandbox with proofs](https://codesandbox.io/s/sleepy-shadow-b9nf4?file=/src/App.js)

zerobias (approved):
All good, big thanks 👍 👍

PR #434, opened 2020-12-17
Add infer helpers for Event<Payload> and Unit<Value>

rymarchikbot:


zerobias:
Cool, thanks! 👍

zerobias:
> with new `UnitValue<T>` we can now simplify some other types
> 
> ```
> export type UnitValue<T> = T extends Unit<infer V> ? V : never
> export type StoreValue<T> = UnitValue<T>
> export type EventPayload<T> = UnitValue<T>
> export type EffectParams<FX extends Effect<any, any, any>> = UnitValue<FX>
> export type EffectResult<FX extends Effect<any, any, any>> = UnitValue<FX['doneData']>
> export type EffectError<FX extends Effect<any, any, any>> = UnitValue<FX['failData']>
> ```

I found that `StoreValue` should throw type error when used with events, so this is not an alias, as we previously thought

PR #444, opened 2021-01-24
Refactor sample type tests

doasync:


zerobias (changes requested):
no regressions, please

zerobias (approved):

PR #445, opened 2021-01-24
New `sample` types

doasync:


src/types/__tests__/effector/sample/sample.test.ts
```diff
@@ -93,7 +98,7 @@ describe('explicit generics', () => {
     })
     expect(typecheck).toMatchInlineSnapshot(`
       "
-      no errors
+      Expected 2 type arguments, but got 1.
```
zerobias, line 101: regression

doasync: I believe I can solve it)

zerobias:
Fantastic work! This going to be a huge improvement in developer experience 🚀

PR #447, opened 2021-01-27
New guard types

doasync:
like sample types but guard :)

PR #452, opened 2021-02-01
Improve sample types

doasync:
fix target unit replacement,
fix clock without source,
fix IfAssignable helper,
fix type inference of events returned from map/filter/merge

### DIFF
--- a/packages/effector/index.d.ts
+++ b/packages/effector/index.d.ts
@@ -109,25 +109,6 @@ export type CompositeName = {
   path: Array<string>
 }
 
-export interface Event<Payload> extends Unit<Payload> {
-  (payload: Payload): Payload
-  (this: Payload extends void ? void : `Error: Expected 1 argument, but got 0`, payload?: Payload): void
-  watch(watcher: (payload: Payload) => any): Subscription
-  map<T>(fn: (payload: Payload) => T): Event<T>
-  filter<T extends Payload>(config: {
-    fn(payload: Payload): payload is T
-  }): Event<T>
-  filter(config: {fn(payload: Payload): boolean}): Event<Payload>
-  filterMap<T>(fn: (payload: Payload) => T | undefined): Event<T>
-  prepend<Before = void>(fn: (_: Before) => Payload): Event<Before>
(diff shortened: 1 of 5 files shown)


src/types/__tests__/effector/sample/sample.test.ts
```diff
@@ -158,7 +300,14 @@ test('generic edge cases', () => {
   }
   expect(typecheck).toMatchInlineSnapshot(`
     "
-    no errors
+    Type 'Store<[any]> | Store<[any, ...any[]]> | Event<any>' is not assignable to type 'Store<A>'.
```
zerobias, line 303: regression

zerobias (changes requested):
we need types for guard more than another pr for sample, especially the one with new regression

src/types/__tests__/effector/sample/sample.test.ts
doasync: Yep, I'm aware of this. The error appeared again after fixing the types for clock without source. I've tried to fix it, but so far without success.

zerobias: That's a task for further releases, not for `21.8.0`

doasync: done

doasync:
The regression is fixed. Here are types for guard: https://github.com/effector/effector/pull/447

doasync:
I've removed explicit generics validation from this PR. It's too complicated and can cause unexpected errors.

packages/effector/index.d.ts
```diff
@@ -730,275 +720,225 @@ type GetResultSCF<S, C, F extends AnyFn> = S extends Store<any> | Combinable
...
   fn: F,
-  target: MultiTarget<ReturnType<F>, T, X>
-  name?: string
+  target: MultiTarget<T, ReturnType<F>>,
+  name?: string,
   greedy?: boolean
 }): T
 
-// Last overload
+// sample's last overload (for readable error messages)
```
zerobias, line 914: 💯

zerobias (approved):
Great work, thanks! 👍

PR #454, opened 2021-02-08
unknown payload in events can be optional (proposal)

doasync:


src/types/__tests__/effector/event.test.ts
```diff
@@ -2,8 +2,8 @@
 import {
   createEvent,
   Event,
-  /*::type*/ CompositeName,
```
zerobias, line 5: Oh, special thanks for removing such old things 😅

igorkamyshev:
Hi, folks. Any updates here? It looks like a great enchantment.

zerobias:
> Hi, folks. Any updates here? It looks like a great enchantment.

Sorry, forgot about this one, actually it is ready for release

PR #456, opened 2021-02-22
New forward types

doasync:


sergeysova:
@doasync is it still actual?

zerobias:
> @doasync is it still actual?

We faced very nuanced issues, so I think this PR is not going to be merged although I always use even closed PR's of @doasync as a knowledge base of typescript tricks

sergeysova:
What about closing this PR?

PR #463, opened 2021-03-11
extend combine shape type

uulaal:
If type of shape is passed explicitly to `combine`, it's impossible to use Stores in it, because of TS error:
image

This extension makes this behavior possible: 
image


zerobias:
Wow, I wonder how we missed such a case. Big thanks! 👍 👍

PR #464, opened 2021-03-12
prevent infinite recursion in `UnionToStoresUnion` type

uulaal:
Fix for #463 

Current type has leak for infinite recursion because of `T extends any` condition - it's always true. But TS still handles this type without any issues, that's why I didn't notice the leak.

New condition has safe guard from infinite recursion and it works the same.

### DIFF
--- a/packages/effector/index.d.ts
+++ b/packages/effector/index.d.ts
@@ -1344,15 +1344,17 @@ export function attach<
 type UnionToStoresUnion<T> = (T extends never
   ? never
   : () => T) extends infer U
-  ? U extends () => infer S
-    ? UnionToStoresUnion<Exclude<T, S>> | Store<S>
+    ? U extends () => infer S
+      ? UnionToStoresUnion<Exclude<T, S>> | Store<S>
+      : never
     : never
-  : never
 type CombineState<State> = {
   [K in keyof State]:
-  | State[K]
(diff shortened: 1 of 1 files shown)


uulaal:
Btw, there's still case to handle: optional parameters of union types:

image

If parameter is required or it's type in store is single, everything's ok: 

image
image

uulaal:
The workaround is to exclude `undefined` here: 
image

It'll be the right solution because, if I'm not mistaken, stores can't have `undefined` state.

zerobias:
Would you add support 

> The workaround is to exclude `undefined` here:
> image
> 
> It'll be the right solution because, if I'm not mistaken, stores can't have `undefined` state.

Would you like to commit that as well, please?

uulaal:
Yep

zerobias:
Looks like last change leads to regression, type test is failed https://github.com/effector/effector/pull/464/checks?check_run_id=2096398301#step:5:262
So you have to avoid that in some way. To check type tests by yourself run `yarn test --testPathPattern='types'`

uulaal:
Can't explain why tests failed due to a simple `undefined` exclusion, but here's a ~~ugly~~ working solution

zerobias:
Yeah, type definitions goes wild sometimes, that's why we have type tests

Thanks! 👍

PR #465, opened 2021-03-15
add or empty string, because babel-plugin-tester do not set state

sergeysova:


sergeysova:
Babel tester works correctly. I forgot to fill `babelOptions` fields

PR #472, opened 2021-03-24
Update naming.md

dimensi:
Hi, previous name feels like simple variable with data, not a function. I think it's wrong.


zerobias:
Seems reasonable, thanks 👍

PR #473, opened 2021-03-24
Proposal for reset in createApi

YanLobat:
Should we consider adding possibility to specify reset Event(s) for createApi? Since reset is since part of an API

### DIFF
--- a/src/effector/createApi.ts
+++ b/src/effector/createApi.ts
@@ -4,8 +4,8 @@ import {forIn} from './collection'
 import {getParent} from './getter'
 import {processArgsToConfig} from './config'
 
-export function createApi(...args: [Store<any>, {[key: string]: Function}]) {
-  let [[store, setters], metadata] = processArgsToConfig(args)
+export function createApi(...args: [Store<any>, {[key: string]: Function}, Event<void>]) {
+  let [[store, setters, resetEvent], metadata] = processArgsToConfig(args)
   const result: Record<string, Event<any>> = {}
   forIn(setters, (fn, key) => {
     const event = (result[key] = createEvent(key, {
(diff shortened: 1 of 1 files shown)


zerobias:
Third argument is very bad news, it's essentially a last resort when we need some feature very much and have no choice. Let's consider converging to common "configuration object" pattern instead

sergeysova:
Why you need another field in `createApi` instead of simple `store.reset(event)`

Before:

```tsx
const $source = createStore(0)
$source.reset(event)
const api = createApi($source, {
  increment: (c) => c + 1,
  decrement: (c) => c - 1,
})
```

After:

```tsx
const $source = createStore(0)
const api = createApi(
  $source,
  {
    increment: (c) => c + 1,
    decrement: (c) => c - 1,
  },
  event,
)
```

What is profit?

yumauri:
I guess point is that you will have all API under one namespace `api`.

What if just always add `reset` to result (if it wasn't defined explicitly)? 🤪

```typescript
export function createApi(...args: [Store<any>, {[key: string]: Function}]) {
  let [[store, setters], metadata] = processArgsToConfig(args)
  const result: Record<string, Event<any>> = {}
  forIn(setters, (fn, key) => {
    const event = (result[key] = createEvent(key, {
      parent: getParent(store),
      config: metadata,
    }))
    store.on(event, fn)
    applyParentHook(store, event)
  })
  
  // ↓↓↓
  if (result.reset == null) {
    const reset = (result.reset = createEvent('reset', {
      parent: getParent(store),
      config: metadata,
    }))
    store.reset(reset)
    applyParentHook(store, event)
  }
  // ↑↑↑

  return result
}
```

PR #479, opened 2021-04-08
Event type fix

doasync:
Here is a reproduction of the issue: [typescript playground](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgZRtApgGjgYyugQxnVQzgF84AzKCEOAcnSqvRzSgYCgv0APSLDgwAnmHRwAKgQBGAG3QAFAlAIgAznAC8iLnDhgCAc3QAuOADsAriBnooAbj1w5oYDHPXb9p5QBkcABKbNAAJgA86jBQwBZG2AQWIgB8Tlw4EBZRcAAkAG4Eclbomjp4hMSk+AAUCIYm5gAM2K4g7k3YMqoWoQCSoeYAjOQAlGkZWfDEUeZV6OHS8koqaurJ2rkFRSUOQA). 

I guess, we cannot assign one store type to another because the events in the `.updates` fields have call signatures... and Typescript is having [trouble](https://github.com/microsoft/TypeScript/issues/35162) working with them. TS checks if `TableParams` is assignable to object, but it should do the exact opposite.

I wrote a [test](https://github.com/effector/effector/compare/event-type-fix?expand=1#diff-73ebfc9702dc74a9c486d1a09493e552f3a60d6f91d37fa20c5dbe7bb8fbb87fR288-R294), and I think I found an interesting [solution](https://github.com/effector/effector/compare/event-type-fix?expand=1#diff-511df50df4b48edff6fd7944a6536f66766e4d3fa4d914c6cdb62149ae4a8965R122-R132).

There are some places to pay attention to: [1](https://github.com/effector/effector/pull/479/files#diff-1776bfbf8be28df8fd961369c42ea61c67e9cdb08ce3d77860ea0d8826f00fb2R311), [2](https://github.com/effector/effector/pull/479/files#diff-f0c3844cb20fa4863948134bef51f78e4414c188d1d1b1ca07d28cf838abda88R248) and [3](https://github.com/effector/effector/pull/479/files#diff-f0c3844cb20fa4863948134bef51f78e4414c188d1d1b1ca07d28cf838abda88L293):
1) `filterMap` returns a correct event type (needs to be updated)
2) Some sort of an edge case with `unknown`... it works as expected with `(event: number) => number`
3) No longer an issue (can be removed)

### DIFF
--- a/packages/effector/index.d.ts
+++ b/packages/effector/index.d.ts
@@ -120,14 +120,16 @@ export type CompositeName = {
 type EventAsReturnType<Payload> = any extends Payload ? Event<Payload> : never
 
 export interface Event<Payload> extends Unit<Payload> {
-  (payload: Payload): Payload
-  (this: Payload extends void ? void : `Error: Expected 1 argument, but got 0`, payload?: Payload): void
+  <T extends Payload>(payload: Payload): T
+  <T extends Payload>(this: Payload extends void ? void : `Error: Expected 1 argument, but got 0`, payload?: T): void
   watch(watcher: (payload: Payload) => any): Subscription
   map<T>(fn: (payload: Payload) => T): EventAsReturnType<T>
   filter<T extends Payload>(config: {
     fn(payload: Payload): payload is T
   }): EventAsReturnType<T>
-  filter(config: {fn(payload: Payload): boolean}): EventAsReturnType<Payload>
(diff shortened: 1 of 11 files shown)


doasync:
According to the results of the discussion, the experiment was considered unsuccessful.

PR #480, opened 2021-04-09
fixed bug with useGate position related to useStore and useVModel

ilajosmanov:
thanks, @Spoki4 for reproduce bug with useGate
WIP: support vue 3 options api. Need to:
- write docs
- types


zerobias:
Thanks! Feel free to open new PR to add docs and types

PR #483, opened 2021-04-25
Fix docs index page horizontal scroll

kelreel:
Removed horizontal scrolling from the main page of the site from the "Features" block (Chrome 89, Firefox). There is probably a better way to solve the problem.

image


zerobias:
Nice, thanks 👍

PR #495, opened 2021-05-30
Specify Observable type in fromObservable

igorkamyshev:
Hello! I have something like this in Aviasales codebase:

```ts
const observable = new Subject<ReallyComplexType>();

const event = fromObservable(observable); // Event<unknown>
```

Sure, I can fix it like this:
```ts
const observable = new Subject<ReallyComplexType>();

const event = fromObservable<ReallyComplexType>(observable); // Event<ReallyComplexType >
```

But it looks too verbose for me. Maybe we can add more specific type to `fromObservable`?


igorkamyshev:
My bad, it doesn't work. I will search for another solution.

zerobias:
But it was supposed to work. This is definitely an issue with the effector types

PR #501, opened 2021-06-23
Fix useEvent typings for void payload

### LINKED ISSUE #500: Typings of useEvent are incorrect
## Missing/Incorrect APIs

The result of `someDomain.createEvent()` is assignable to `() => void`, but `useEvent(event)` is not:
```
TS2322: Type '(payload: void) => void' is not assignable to type '() => void'.
```

igorkamyshev:
Hello! My team-mate found `useEvent` types issue and I wanna to fix it. I am not so good at types writing, so please review my PR and give me feedback.

Fix #500

### DIFF
--- a/packages/effector-react/index.d.ts
+++ b/packages/effector-react/index.d.ts
@@ -101,7 +101,9 @@ export function createReactState<
   Com extends React.ComponentType<any>
 >(store: Store<State>, Component: Com): React.ComponentType<State>
 
+export function useEvent(event: Event<void>): () => void
 export function useEvent<T>(event: Event<T>): (payload: T) => T
+export function useEvent<R>(fx: Effect<void, R, any>): () => Promise<R>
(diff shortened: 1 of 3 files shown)


zerobias (changes requested):
Current tests are passed, so it's probably ok, but its better to add new type tests with your case to be sure that it will work as expected. Tests for useEvent are [here](https://github.com/effector/effector/blob/master/src/types/__tests__/effector-react/effectorReact.test.tsx#L55).



Contribution guide for type tests: https://github.com/effector/effector/blob/master/CONTRIBUTING.md#type-tests

Feel free to write to me if you have any questions

igorkamyshev:
Sorry, you are right. Fixed ☑️

zerobias (approved):

zerobias:
Great, thanks! 👍

PR #504, opened 2021-07-12
Add the missing image visualization of sample

c01nd01r:


zerobias:
Big thanks! 👍 I didn't even think it would be that easy 😄

PR #506, opened 2021-07-22
change requestsSend to requestsSent

aldibatyr:
grammatically and semantically correct naming for this store would be `requestsSent` since it counts how many requests have been sent so far.


igorkamyshev:
Thanks, you're right

PR #520, opened 2021-09-02
deprecate `filter`, add `except`

5angel:


igorkamyshev:
In my opinion, we must not deprecate anything right after a major release. We have to care about users and their applications.

src/effector/__tests__/event/except.test.ts
```diff
@@ -0,0 +1,26 @@
...
+    event(2)
+    event(-1)
+    expect(argumentHistory(fn)).toEqual([2])
+  })
+  it('works with undefined as well', () => {
+    const fn = jest.fn()
+    const event = createEvent()
+    const filtered = event.filter({
+      fn: x => x === undefined,
+    })
```
AlexanderKhoroshikhRMR, line 19: Probably, it should be except, instead of filter?
```suggestion
    const filtered = event.except(x => x === undefined)
```

AlexandrHoroshih, line 19: Probably, it was meant to be `except` here and not `filter`?
```suggestion
    const filtered = event.except( x => x === undefined)
```

5angel:
closing due to the issue being up to debate

PR #524, opened 2021-09-18
docs: add slots for react recipe

binjospookie:


zerobias:
Very interesting idea, thanks 👍

PR #525, opened 2021-09-20
Update slots.md

binjospookie:
I found how to improve DX. In fact, we no need to return $slot. We can return Slot component.
It makes integration simple and more type-stronger code.


binjospookie:
https://github.com/space307/effector-react-slots/commit/ec2ce2e5eeda71d6a28ed92d22f877172ed952ac#diff-3c6e29a9bf3e4efab86c55f7d4034d1366e349e6f477a02eb43b57316d350c09L3
as proof :)

zerobias:
👍 👍

PR #543, opened 2021-10-28
just wana try out crazy stuff

YanLobat:
almost no one should observe these changes

@zerobias I was wondering how the fn is processed? My curiosity was about an ability reducing several functions (in these monkey changes I suggest applying fn to our source&clock and having an ability to filter fn result just after)

know the thing about the types but in terms of just js code will this idea kinda work?


YanLobat:
@zerobias by the way check this out maybe you'll find some references around https://blazegraph.com/

Issue #278, opened 2019-12-22
Why uses Effector?


r3nya:
Hi folks! 
Curious to know who is using Effector! It would be really appreciated if you:

1. Share your company name
1. Share your logo (in which case it might be used in the readme/frontpage/etc)
1. What kind of project? How large? (codebase size or team size)
1. If your project is open source or the public, please share a link!

Thanks in advance! ❤️

zerobias:
Here @sergeysova talking about effector in REDMADROBOT

https://soundcloud.com/5minreact/063-effector 🇷🇺

Tauka:
1. Okoo
2. ![Default](https://user-images.githubusercontent.com/15087247/71434120-3413ea80-270d-11ea-9398-409cfcc33326.png)
3. Educational startup in Kazakhstan, team: 7 people, LOC: 10k

amabunny:
1. I've used it pretty good in enterprise company.
Can't call the name because it's privacy data.

2. *Logo was kidnapped by UFO*

It working so good to store temporary data and solves a lot a problem that comes with state managing in large enterprise apps because it's simple to read and understand. I want to notice that time I was usually spending to write a lot of code was replaced with time to design a better architecture for my app and it's awesome! 

I like this tool so much and want to use it in my main stack everytime.

3. Its a CRM-system for sales with big business data like customers base, lot of mobile devices base and strong forms with additional computation logic. Want to say that it looks really good with big forms. Our team is about three people with next "Junior-Middle-Senior" range.

P.S. Still waiting devtools so much!

Regards, Sergei.

ivanov-v:
1. [STM Labs](http://stm-labs.ru/)
2. ![logo2](https://user-images.githubusercontent.com/13759065/73070727-25faea80-3ec2-11ea-972a-8cfaedd33621.png)

egorguscha:
[automation hero](https://automationhero.ai/)
![download](https://user-images.githubusercontent.com/22044607/75612380-e44e0700-5b33-11ea-8fe7-fc460ac4561d.png)

sergeysova:
[REDMADROBOT](http://redmadrobot.com/)
![image](https://user-images.githubusercontent.com/5620073/82443193-c0062480-9aa9-11ea-89c2-96f7b7118927.png)

MirzayevFarid:
- We’re using Effector at [Radity](https://radity.com/)

[...]

PR #547, opened 2021-10-30
#148 Node.js AsyncLocalStorage example

darky:
Example how we use Effector with Node.js + AsyncLocalStorage

### DIFF
--- a/docs/recipes/nodejs-async-local-storage/example.md
+++ b/docs/recipes/nodejs-async-local-storage/example.md
@@ -1,0 +1,110 @@
+---
+id: nodejs-async-storage-example
+title: Node.js AsyncLocalStorage example
+sidebar_label: Node.js AsyncLocalStorage example
+---
+
+The following example is a Node.js counter that stores data to Redis.
+It uses store, events and effects, which will be created from scratch and saved to AsyncLocalStorage for each HTTP request.
+
+
+```typescript
+import { attach, createEffect, createEvent, createStore } from 'effector';
+import { diDep, diInit, diSet } from 'ts-fp-di';
+import { diEffector } from 'ts-fp-di-effector';
+import Koa from 'koa';
+import Router from 'koa-router';
+import Redis from 'ioredis';
+
+const diEff = diEffector({
+  onCreateEffect(sid, effect) {
+    effect.watch((val) => {
+      console.log(`Effect "${sid}" call with value: ${val}`);
+    });
+    effect.doneData.watch((val) => {
+      console.log(`Effect "${sid}" done with value: ${val}`);
+    });
+    effect.failData.watch((val) => {
+      console.log(`Effect "${sid}" fail with value: ${val}`);
+    });
+  },
+  onCreateEvent(sid, event) {
+    event.watch((val) => {
+      console.log(`Event "${sid}" call with value: ${val}`);
+    });
+  },
+  onCreateStore(sid, store) {
+    store.watch((state, val) => {
+      console.log(`Store "${sid}" mutation with value: ${val}`);
+      console.log(`Store "${sid}" mutation, current state: ${state}`);
+    });
+  },
+});
+
+const increment = diEff('increment', () => createEvent());
+const decrement = diEff('decrement', () => createEvent());
+const reset = diEff('reset', () => createEvent());
+
+const pullCounterFx = diEff('pullCounterFx', () =>
+  createEffect<void, number>(async () => {
+    const login = diDep('login');
+    const count = await redis.get(`${login}:counter`);
+    return Number(count ?? 0);
+  })
+);
+
+const pushCounterFx = diEff('pushCounterFx', () =>
+  attach({
+    source: $counter(),
+    effect: createEffect<number, number>(async (count) => {
+      const login = diDep('login');
+      await redis.set(`${login}:counter`, count);
+      return count;
+    }),
+  })
+);
+
+const $counter = diEff('$counter', () =>
+  createStore(0)
+    .on(increment(), (state) => state + 1)
+    .on(decrement(), (state) => state - 1)
+    .on(pullCounterFx().doneData, (_, value) => value)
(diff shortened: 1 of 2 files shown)


sergeysova:
@darky Thank you for this example!

But, why you need DI there?

darky:
Yep, I use in example our tiny wrappers around *AsyncLocalStorage* - `ts-fp-di`, `ts-fp-di-effector`

This utils help you to bootstrap development process in proper way with Effector on Node.js with less hassle and much more clean code.

Little overhead here, that `diEff` factory required for define our **Units**
But then you can use **Stores**, **Events** or **Effects** "as is" in transparent way and it will be "magically" consider your current application lifecycle (current HTTP request, handle specific MQ message and whatever). It's like "mini-app", like Scope with specific **Events**, **Effects** or **Stores** that live in sandbox and isolated from each other.

igorkamyshev:
In my opinion, this example will be look better in documentation for DI-library. So, maybe we should put the link to this library to awesome list.

zerobias:
I looked after ts-fp-di for some time, interesting idea in general but I concern that it directly interferes with main effector’s concept — static (mean one-time) initialization of all units in application. This is what makes effector the effector

sergeysova:
I think in this example DI can be replace with the Fork API

sergeysova:
Something like this https://gist.github.com/sergeysova/289a6948a468447c21c9a1815a9e667c/revisions

But the most important thing is in the **revisions tab**

darky:
> In my opinion, this example will be look better in documentation for DI-library. So, maybe we should put the link to this library to awesome list.

Agree, link in awesome list LGTM

> I looked after ts-fp-di development for some time, interesting idea in general but I concern that it directly interferes with main effector’s concept — static (mean one-time) initialization of all units in application. This is what makes effector the effector

Yep, it another concept. Here units allocated dynamically for each lifecycle on backend (HTTP request and so on). I don't try to sell this idea as main for Node.js development with Effector, but want pay attention of community as alternative way.

> I think in this example DI can be replace with the Fork API

Even we start develop with Fork API, but don't like it via multiple reasons:
 
1. We consider, that fork "whole world" on each HTTP request has high cost. How to scale it? We work on very large project (https://github.com/effector/effector/issues/278#issuecomment-754028333) with thousands (maybe ten thousands) units in potential. We think, that in future this will block event loop in Node.js and drastically decrease performance of our backend. From the other side, init batch of specific units for specific HTTP request look good to us.

2. Out investigation of Domain API led to the fact that it is a few fragile:

* https://effector.dev/docs/api/effector/scope#imperative-effects-calls-with-scope
* https://effector.dev/docs/api/effector/scopeBind

3. We don't want explicit work with `scope` object, pass it to any place, consider it in code. We just want to use **Stores**, **Events** or **Effects** "as is" and context will be considered automatically.

> Something like this https://gist.github.com/sergeysova/289a6948a468447c21c9a1815a9e667c/revisions

Thanks for example! I run it. Sorry, but logging don't work, seems mistake occured. Pass `domain` to `fork` method don't help.

zerobias:
@darky performance issues have been fixed in version 22, `fork` performance has been improved by 100x. At this time, creating units per request will be tens of times slower

zerobias:
And as for fragility - fork was implemented based on ideas from IndexedDB transactions and AsyncLocalStorage contexts, so it works with _exactly_ the same limitations

zerobias:
@darky here is your example with Koa online:
- [https://fork-with-nodejs-kkr4026ru8so.runkit.sh/__increment__?_login_=_darky_ (published endpoint)](https://fork-with-nodejs-kkr4026ru8so.runkit.sh/increment?login=darky)
- [https://runkit.com/zerobias/fork-with-nodejs (source code)](https://runkit.com/zerobias/fork-with-nodejs)

darky:
> performance issues have been fixed in version 22, fork performance has been improved by 100x. At this time, creating units per request will be tens of times slower

@zerobias thanks you very much for this awesome work! Really, we think about it, when Effector was 21 version. Seems event loop stuck problem can be delayed to even further. For good here need simulate stress testing with thousands units. I think about it.

> @darky here is your example with Koa online

Thanks! I run it locally and logging not working as is here https://gist.github.com/sergeysova/289a6948a468447c21c9a1815a9e667c/revisions

sergeysova:
@darky 
> Thanks! I run it locally and logging not working as is here

You need to use babel and [effector/babel-plugin](https://effector.dev/docs/api/effector/babel-plugin/) to automatically set `sid` to each unit. That's why my example doesn't work without it.

You also can set `sid` manually as @zerobias shows

darky:
@sergeysova set `sid` manually don't help.
@sergeysova or @zerobias can you please update your gist for working logging?

You can simply run my example and logging will be working like a charm.
I use this as :heavy_plus_sign: 1 point to `diEff` factory, it guarantee that all units will be catched and used for logging.
Logging matter on backend, you can't reproduce bug that happens 3 days ago on particular user in particular situation without logging.

zerobias:
@darky sorry, forgot to _actually_ use the domain, example is fixed 👌 

[source code](https://runkit.com/zerobias/fork-with-nodejs/2.0.0)

real usage logs:
image

darky:
@zerobias thanks, now logs works. But, sorry for dissatisfaction.
`pushCounterFx`, `pullCounterFx` missing in logs.
How to log effects, created by `attach` method? Or maybe stores, created by `combine` method? And so on.

zerobias:
@darky for more complex cases I suggest to look at [patronum debug](https://github.com/effector/patronum#debug)

darky:
> for more complex cases I suggest to look at patronum debug

Ok, it's actually that `diEff` already do - wrap unit with custom function. 

Thanks you guys for answers and time spent. No more bother with "hipster" AsyncLocalStorage Domain.

sergeysova:
@darky Thank you for your case with DI!

PR #585, opened 2021-12-21
Fix implicit any in filter

doasync:


zerobias (changes requested):
Very impressive work, I was about to give up. Could you also handle cases with loss of tuples, please?

zerobias (approved):
I just realized that issue was here before your changes, sorry, my bad! I wonder how critical these issues really are, but I think we could learn it in prod

Very big thanks, as I said, I was just about to give up on this one 🙏

PR #588, opened 2021-12-23
Fix wide tuple (sample types)

doasync:
The wide tuple check is built into the `IfAssignable` helper:
```ts
      : T extends Array<any>
        ? number extends T['length']
          ? N
          : U extends Array<any>
            ? number extends U['length']
              ? N
              : TupleObject<T> extends TupleObject<U> ? Y : N
            : N
        : N
```


zerobias (approved):
Wow! I was tinkering with that feature, but didn’t advanced at all 😃 Thanks! 👍👍

PR #589, opened 2021-12-24
Update createGate.ts

dzhipef:
[https://stackblitz.com/edit/react-gesa77?file=src/App.js](problem)


zerobias (changes requested):
Could you add [a test](https://github.com/effector/effector/blob/master/src/react/__tests__/createGate.test.tsx) that shows the problem specifically in Gate please? I understand what's being done technically in the PR, but I don't see exactly what problem it solves.

PR #591, opened 2021-12-27
Filter falsy values with Boolean constructor

doasync:


zerobias:
Useful improvement, thanks! 👍 👍

PR #592, opened 2021-12-27
Fix sample generic edge case

doasync:
Merge after https://github.com/effector/effector/pull/591 to avoid conflicts (PR target will automatically change to `master`)

### DIFF
--- a/packages/effector/index.d.ts
+++ b/packages/effector/index.d.ts
@@ -1701,7 +1701,7 @@ type SampleFilterTargetDef<
         : [message: {error: 'function should accept data source types'; got: FN}]
     : FLBool extends BooleanConstructor
       ? Mode extends Mode_Flt_Fn_Trg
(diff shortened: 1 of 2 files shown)


packages/effector/index.d.ts
```diff
@@ -1701,7 +1701,7 @@ type SampleFilterTargetDef<
         : [message: {error: 'function should accept data source types'; got: FN}]
     : FLBool extends BooleanConstructor
       ? Mode extends Mode_Flt_Fn_Trg
-        ? FNAltArg extends (arg: infer FNArg) => any
+        ? FNAltArg extends (arg?: any, clk?: any) => any
```
zerobias, line 1704: Great trick, need to try to use it in other places in types, it should be much easier to use than checking functions case-by-case

zerobias:
Splendid! Thanks 👍 👍

zerobias (approved):

PR #593, opened 2021-12-28
Do not infer type from returned unit for sample

doasync:
It's interesting that it became possible at all with this approach. We could use this feature for `prepend` (but we don't need it for sample)


zerobias (commented):
interesting, inference order looks like a feature with a big blast radius; are there any other implications here? how do types in nested inlined operator calls behave?

zerobias:
I think that this improvement should have no bad consequences, thanks!

PR #596, opened 2022-01-06
Fix Recipes -> React -> TODO creator

romadryud:
When I tried to apply this example in life, typescript pointed out several typing errors, so this PR will fix them. Also, I make updates for example. I know that I should have created an issue first, but I thought that there are a small number of changes and if these edits are not accepted, I won't be particularly upset, but I remain positive :)

Thanks for the great project.


zerobias:
Thanks! 👍 👍

PR #625, opened 2022-02-06
Support store as `allSettled` argument

### LINKED ISSUE #608: `allSettled` doesn't support `$store` as argument silently

If we provide any store to `allSettled` it will change it's value, but later `serialize` won't detect it.

https://share.effector.dev/wcQz3AOQ

Throw exception on passing store to `allSettled` or fully support it.

`master`

AlexandrHoroshih:
Fixes #608 (by fixing serialize issue)

Changelog:
- added tests for `allSettled($store, ...)` cases
- added fix for the case

Apparently, the `getParent(storeStack)` line in `storeChange` node defintion is needed to check, if store is created from combine or map, but the same check also led to `$store` as starting unit for `allSettled` being skipped in serialization - because there is no "parent trigger" 🤔

### DIFF
--- a/src/effector/__tests__/fork/allSettled.test.ts
+++ b/src/effector/__tests__/fork/allSettled.test.ts
@@ -7,6 +7,7 @@ import {
   allSettled,
   sample,
   serialize,
+  createStore,
 } from 'effector'
 
 test('allSettled first argument validation', async () => {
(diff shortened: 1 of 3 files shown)


zerobias (approved):

zerobias:
Thanks! 👍

PR #629, opened 2022-02-12
added: edge case for split

### LINKED ISSUE #628: TypeScript errors using split from effector@22.2.0

TypeScript errors using split from effector@22.2.0

A minimal [demo](https://www.typescriptlang.org/play?ts=4.4.4#code/JYWwDg9gTgLgBAbwM5gDbBgGjgYygUwEMZ8BRAN3wDstcDj8BlGafbJQ8VfAXzgDMoEEHADk+fv3w4WUUQCh5OCFSTwAJHAC8dIiWasAPFQCuIAEb4ocAD5xTqVAD4AFA9QBKJSrVx85eB08PTJKGmMzS2s7d1cvZVV4AHcoDFDA3QYKahhDBDhCAC57SKtsc2LTCys4HjiAbkUUdBgXBHk4TtxUCBwAa2KOLnw2jq7xnB7+4v8sMfHOpAgTKBx8YvVMeYX+KmKXdXJCVBM2WYAFQgBPHsIAEw9tJzg2go2jk7Y4Cr8Ay5uIPdal5xjwPFtxiBiDgABbFdoLLopNL7QiYcyPLTPQhwACEWh07jgADJid88QSSo5tjwIV0cIQkPgkPDtuNkSRihz0nTOjx5GCgA) of the problem.

No TypeScript errors

effector@22.2.0
typescript: 4.4.4 (playground). I have the same problem in my project under TypeScript 4.5.2

binjospookie:
I'm not sure but we probably need to check that  `UnitValue<Match> extends CaseTypeReader<Cases, keyof Cases>` in the first SplitType fork.

I doubt it because I see the failed tests. Maybe that's how it was supposed to be

### DIFF
--- a/packages/effector/index.d.ts
+++ b/packages/effector/index.d.ts
@@ -731,13 +731,12 @@ type SplitType<
 
       : {error: 'not implemented'}
 
-  : UnitValue<Match> extends CaseTypeReader<Cases, keyof Cases>
-    ? Config
-    : {
-        error: 'source type should extends cases'
-        sourceType: UnitValue<Source>
-        caseType: CaseTypeReader<Cases, keyof Cases>
-      }
+  : {
+    error: 'source type should extends cases'
+    sourceType: UnitValue<Source>
+    caseType: CaseTypeReader<Cases, keyof Cases>
+  }
+
 /**
  * Chooses one of cases by given conditions. It "splits" source unit into several targets, which fires when payload matches their conditions.
  * Works like pattern matching for payload values and external units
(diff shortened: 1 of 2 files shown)


zerobias (changes requested):
Seems like this change just disable typechecking [for certain cases](https://github.com/effector/effector/runs/5167931284?check_suite_focus=true#step:5:1571) thereby passing its own tests
Of course, current behavior is not how things assumed to work, the only catch is that every change in typings should care about _every_ edge case of the method

I suggest to add tests via this PR as they are important to realize the edge case and try to fix it later; I don’t think this going to be solved rapidly

binjospookie:
Reverted types changes.

zerobias:
Thanks!

PR #630, opened 2022-02-13
Warn about missing sids during serialize

### LINKED ISSUE #567: Effector does not warn, if store without sid is used as a value in fork
Store without sid still can be used as a value in fork, even though it leads to weird behaviour without any warnings or errors

REPL: https://share.effector.dev/35CeFqED

Effector warns or errors, if store without sid is used in fork values, so user can check babel-plugin usage or provide sid's manually

`effector@22.0.0` and up

AlexandrHoroshih:
Closes #567 

**Changelog**
- Added tests for the problem
- Added check at `serialize`, if serialization of this scope cannot be performed in a reliable way and should be warned
- Warning for every scope is set by separate node, which is scheduled based on store's meta - like with `storeChange`
- Added caution note in the `serialize` docs about sids

### DIFF
--- a/docs/api/effector/serialize.md
+++ b/docs/api/effector/serialize.md
@@ -20,6 +20,10 @@ A companion method for [_fork_](./fork.md). Allows to get a serialized value for
 
 An object with store values using sids as a keys
 
+:::caution
+If a store [does not have a sid](./babel-plugin.md#sid), its value will be omitted during serialization.
+:::
+
 ### Example
 
 Serialize forked instance state
--- a/website/client/i18n/ru/docusaurus-plugin-content-docs/current/api/effector/serialize.md
+++ b/website/client/i18n/ru/docusaurus-plugin-content-docs/current/api/effector/serialize.md
@@ -20,6 +20,10 @@ A companion method for [_fork_](./fork.md). Allows to get a serialized value for
 
 An object with store values using sids as a keys
 
+:::caution
+If a store [does not have a sid](./babel-plugin.md#sid), its value will be omitted during serialization.
+:::
+
 ### Example
 
 Serialize forked instance state


src/effector/__tests__/fork/serialize.test.ts
```diff
@@ -455,3 +455,70 @@ test('onlyChanges: false supported only in domain-based scopes', () => {
...
+    console.error = consoleError
+  })
+  test('serialize: doesn not warn on mapped or combined stores', () => {
+    const error = jest.fn()
+    const consoleError = console.error
+    console.error = error
+
+    const $store = createStore('value')
+    const $mapped = $store.map(s => s)
+    const $combine = combine($store, $mapped, (_, m) => m)
```
zerobias, line 509: `combine` does have sid here, usually in tests it could be hidden like that:

```ts
const $combine = {_: combine}._($store, $mapped, (_, m) => m)
```

src/effector/fork/serialize.ts
```diff
@@ -11,6 +11,9 @@ export function serialize(
   scope: Scope,
   config: {ignore?: Array<Store<any>>; onlyChanges?: boolean} = {},
 ) {
+  if (scope.warnSerialize) {
+    console.error("provided scope cannot be serialized in a reliable way because some stores are missing sid. Please, check that babel-plugin is working or provide sid manually");
```
zerobias, line 15: Please reduce the length of the text. The message should fit into one neat sentence.
Libraries like react have a huge dev bundle with obscure error messages in production, nevertheless, the effector has only one bundle, so must keep track of the size of it

src/effector/__tests__/fork/serialize.test.ts
```diff
@@ -455,3 +455,70 @@ test('onlyChanges: false supported only in domain-based scopes', () => {
...
+
+    allSettled($store, {scope, params: 'scope value'})
+
+    const result = serialize(scope)
+    expect(result).toEqual({
+      [$store.sid as string]: 'scope value',
+    })
+    expect(scope.getState($store)).toEqual('scope value')
+    expect(error).toHaveBeenCalledTimes(0)
+    console.error = consoleError
```
zerobias, line 500: All interactions with environment should be performed in beforeEach/afterEach, otherwise when a single test will fail, it will broke console for rest of the suite

```ts
describe('serialize: missing sids', () => {
  let fn
  let consoleError
  beforeEach(() => {
    fn = jest.fn()
    consoleError = console.error
  })
  afterEach(() => {
    console.error = consoleError
  })
})
```

zerobias (changes requested):
Will it work with effects? They have internal stores. Add tests for that, please

src/effector/createEffect.ts
```diff
@@ -133,7 +133,11 @@ export function createEffect<Payload, Done, Fail = Error>(
     return req.req
   }
 
-  const inFlight = (instance.inFlight = createStore(0, {named: 'inFlight'})
+  const inFlight = (instance.inFlight = createStore(0, {
+    named: 'inFlight',
+    // @ts-expect-error
+    serialize: 'ignore',
```
AlexandrHoroshih, line 139: I think, that this case is similiar to internal stores of the libraries: if the store is not meant to be transferred over the network, then it is must have `serialize: 'ignore'` set 🤔

docs/api/effector/serialize.md
```diff
@@ -20,6 +20,10 @@ A companion method for [_fork_](./fork.md). Allows to get a serialized value for
 
 An object with store values using sids as a keys
 
+:::caution
+If a store [does not have a sid](./babel-plugin.md#sid), its value will be omitted during serialization.
+:::
```
AlexandrHoroshih, line 25: image

Issue #563, opened 2021-11-24
Deprecate derived units in `target`


AlexandrHoroshih:
**What is the current behavior:**
Usage of `.on` method is deprecated on derived stores and to be removed in the next release - this is very good news and, i think, makes everything a lot more consistent and predictable 👍 

But now usage of derived stores as a `target` in operators is not warned or error by effector, so the problem of unexpected changes to derived stores still persists 😢 

```ts
$readOnlyStore = $store.map(v => v)

// yay! error in console!
$readOnlyStore.on(ev, () => {})

// no error :(
sample({
 clock: $store,
 target: $readOnlyStore,
})
```
Also i think that stores created by restore from an event or effect should be considered as derived too, since they also have explicit dependency at the moment of creation: restore(changed, null) 🤔
UPD: Since restored event state still can be used by app and can't be derived from, for e.g., state in serialized scope during SSR, it is definetly not derived (at least, in a sense similiar to mapped stores)

**Please provide the steps to reproduce and if possible a minimal demo of the problem via https://share.effector.dev, https://codesandbox.io or similar**

https://share.effector.dev/yqvd359F

**What is the expected behavior:**
Any attempt of adding a new source of updates to dervied store is forbidden and warning/error is shown.
If such behaviour is needed - it is better to create classic (not derived) store and subscribe it to changes of original derived store.

**Which versions of effector packages, and which browser and OS are affected by this issue? Did this work in previous versions of effector?:**
`effector@22.0.0^`

zerobias:
So we need to add such validations to:
- forward to
- sample target
- guard target
- split cases

Anything else?

sergeysova:
Maybe add validation on the kernel level? To add support for patronum

zerobias:
This is added to the metadata of the graph node anyway, assume something like that:

```js
if (unit.graphite.meta.derived) throw Error('derived unit cannot be used as a target')
```

AlexandrHoroshih:
I updated a REPL example a bit:
- dervied events in `target` are also not warned yet
- `merge` returns an not-derived unit and its direct call is not warned

https://share.effector.dev/0f1bnC6c

zerobias:
@AlexandrHoroshih thanks, I wrote more tests based on your findings

sergeysova:
> Also i think that stores created by restore from an event or effect should be considered as derived too, since they also have explicit dependency at the moment of creation: restore(changed, null) 🤔

[...]

PR #635, opened 2022-02-19
fix: controlled units direct usage is warned

AlexandrHoroshih:
Fixes missing part of #563 

There is a public units, which aren't derived, but are controlled by their parent units - like `$store.updates` or `effect.inFlight`.
Their direct usage currently isn't warned by `effector`, but it should be, since otherwise it may lead to confusing behavior.
Even though internally these units are not derived, they are semantically dervied for the external observer, as they fully depend on their parent units, so, i think, these units should be marked as derived too.

**Changelog**
- Added tests on direct usage of controlled units of store and effect (`$store.updates`, `effect.finally`, `effect.inFlight`, etc)
- `effect.inFlight` is exposed as an mapped store, with its parent being hidden inside effect.
- `effect.finally` is marked as derived in its config
- `$store.updates` is marked as derived in its config
- Added caution note about direct calls of `$store.updates`


docs/api/effector/Store.md
```diff
@@ -478,6 +478,10 @@ $store.updates
 
 Use case: watchers, which will not trigger immediately after creation (unlike [_store.watch_](Store.md#watchwatcher))
 
+:::caution Important
+Do not manually call this event. It is event that depends on a store.
```
AlexandrHoroshih, line 482: image

website/client/i18n/ru/docusaurus-plugin-content-docs/current/api/effector/Store.md
```diff
@@ -384,6 +384,10 @@ add(3)
 
 Дочернее событие стора, будет вызвано при каждом обновлении стора. Используется только для определения сайд-эффектов через _store.updates.watch_, где будут срабатывать начиная с первого обновления, в отличии от [_store.watch_](#watch), обработчики в котором запускаются при создании немедленно
 
+:::caution Вызов вручную запрещён
+Это свойство управляется самим стором.
```
AlexandrHoroshih, line 388: image

zerobias:
Amazing, thanks! 🙏

PR #638, opened 2022-03-24
Implement `classList` property for forest

### LINKED ISSUE #599: Add `classList` to `spec()` to allow independently changing CSS-classes
## Problem

Sometimes I need to set CSS-classes for single element from different places. For example:

```ts
function Card(fn) {
  h('div', {
    attr: { class: 'padding-4' },
    fn() {
      fn()
    },
  })
}

Card(() => {
  spec({ attr: { class: 'bg-white' } })
})
```

But the second class attr in the spec won't work.

## Proposal

I propose to add support for [`classList`](https://developer.mozilla.org/en-US/docs/Web/API/Element/classList).

1. Object form of classes, allows to enable/disable class via boolean-store value

```ts
spec({
  classList: { 'size-big': $isBig, white: true }
})
```

2. Array form must allow to enable classes provided by store:

```ts
const $size = createStore('size-big');

spec({
  classList: [$size, 'enabled']
})
```

3. Different calls of spec with `classList` should merge (lower one wins?)

```ts
h('div', () => {
  spec({ attr: { class: 'zero' } })
  spec({ classList: { first: true, second: false } })
  spec({ classList: ['second', 'third'] })
})
```

should be rendered as

```html
<div class="zero first second third"></div>
```

sergeysova:
Closes #599


src/forest/method/h.ts
```diff
@@ -73,7 +88,6 @@ function createPropsOp<T, S>(
       group: leaf.root.leafOps[leaf.fullID].group,
     })
     leaf.root.leafOps[leaf.fullID].group.ops[opID] = op
-    const ctx = initCtx(value, leaf)
```
zerobias, line 76: Was this necessary for something specific? If you call initialization at the beginning, then [some data](https://github.com/effector/effector/pull/638/files#diff-df0554ad39e0813ff09cce9f39ab1f53bd84a9927f11a9309c96c6715941794fR90) may not make it [to processing](https://github.com/effector/effector/blob/0e20c015c0bf488affbaa5afcbebec86ed28e33a/src/forest/method/h.ts#L534)

src/forest/method/h.ts
```diff
@@ -335,6 +352,28 @@ export function h(tag: string, opts?: any) {
...
+            type: 'classList',
+            field: is.unit(property.name)
+              ? property.name
+              : createStore(property.name),
+            value: is.unit(property.enabled)
+              ? property.enabled
+              : createStore(property.enabled),
+          })
+          if (is.unit(property.name)) processStoreRef(property.name)
+          if (is.unit(property.enabled)) processStoreRef(property.enabled)
```
zerobias, line 368:
```suggestion
          const $name = is.unit(property.name)
              ? property.name
              : createStore(property.name)
          const $enabled = is.unit(property.enabled)
              ? property.enabled
              : createStore(property.enabled)
          
          draft.seq.push({
            type: 'classList',
            field: $name,
            value: $enabled,
          })
          processStoreRef($name)
          processStoreRef($enabled)
```
Not sure that this is important, but let's proceed each created store

zerobias (changes requested):
Good job 👍  Please consider make changes in a few places.

Also, tests [are failed](https://github.com/effector/effector/runs/5751540849?check_suite_focus=true#step:6:479) because of the way `classList` works: indeed, when you remove the very last class token, there sill will be an empty `class` attribute. Try to run this code to see:
```js
const div = document.createElement('div')
div.appendChild(new Text('foo'))
document.body.append(div)
const cl = div.classList
cl.add('foo')
cl.remove('foo')
```
You need to choose, whether it should work in that way (and then update your snapshots) or improve implementation to explicitly remove `class` attribute in case of empty list

src/forest/method/h.ts
sergeysova: I had a exception "value used before initialization"
telegram-cloud-photo-size-2-5325802100816460378-y

zerobias:
Great, thanks! 👍 👍

PR #646, opened 2022-04-04
Add new helper method `createWatch`

igorkamyshev:
# What is this

`effector-react` uses `createWatch` to create subscription on particular scope. The PR moves this function to `effector` and allows external consumers to use it.

# Why

Any tool which attempts to connect effector with an external library will reimplement `createWatch`, if it works with `scope`-full runtime. Some cases:
1. `effector/solidjs` strictly requires the same function to implement support of forked scopes.
2. `effector/node-monitor` (the library that I'm going to write to export internal metrics of effector to Node.js monitoring tools) strictly requires the sample function because on server-side it's useless to watch all events — we should restrict it to current scope.

However, internal implementation of `createWatch` is a kind of intimate for effector, so it's better to keep it inside.

# What is inside
+ `createWatch` copy-pasted from `effector-react` to `effector`
+ it's modified to accept any `unit`, not only `stores`
+ the signature modified to maintain simplicity and expandability
+ docs (RU and EN) is added
+ `Unit`-interface in `index.h` is made generic with default value `unknown`

# Let's discuss
+ What shall we do with old implementation of `createWatch` in `effector-react`? We cannot delete it because it'll cause breaking changes. Shall we do it in the major release?


zerobias (approved):
That's a great idea! I think this will make it a lot easier to support ecosystem packages. Will be scheduled for the next minor release

zerobias:
> Shall we do it in the major release?

Yes, `effector-react` implementation will be updated in the next major release

PR #671, opened 2022-04-27
Fix attach operator with plain function case defails in documentation

yialo:
[documentation] `attach` operator has an overload with plain function as `effect` option. Clarify relation of `source` option to `effect` one in this case.


sergeysova (approved):

sergeysova:
@yialo Thank you!

PR #678, opened 2022-05-13
Allow to to set many classes in a single `classList` property

sergeysova:
Previously if I need to switch many classes via single condition I need to list them all:

```ts
const $isEnabled = createStore(true)
const $isDisabled = $isEnabled.map(is => !is)

spec({
  classList: {
    'bg-black': $isEnabled,
    'text-white': $isEnabled,

    'bg-white': $isDisabled,
    'text-black': $isDisabled,
  },
})
```

This PR allows to list many classes in the same property:

```ts
const $isEnabled = createStore(true)
const $isDisabled = $isEnabled.map(is => !is)

spec({
  classList: {
    'bg-black text-white': $isEnabled,
    'bg-white text-black': $isDisabled,
  },
})
```

It is especially useful with tools like Tailwind

---

As a benefit it supports dynamic classes:

```ts
const $classes = $isEnabled.map(is => is ? 'text-white bg-black' : 'text-black bg-white')
spec({ classList: [$classes] })
```

### DIFF
--- a/src/forest/__tests__/classList.test.ts
+++ b/src/forest/__tests__/classList.test.ts
@@ -83,6 +83,25 @@ it('supports setting static object class with class attr', async () => {
       "
     `)
 })
+it('supports setting static list of classes in a single string', async () => {
+  const [s1] = await exec(async () => {
+    using(el, () => {
+      h('div', {
+        text: 'content',
+        classList: ['example another', 'first second'],
+        attr: {class: 'foreign'},
+      })
+    })
+    await act()
+  })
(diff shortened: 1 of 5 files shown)


sergeysova:
> Tests failing only because `Gate` stores don't have sid

zerobias:
> tests failed

That’s why I don’t leave repository in failed state: you either fix an issue rapidly or invert/disable test with a description why that’s happened in that way

zerobias:
Good, thanks! 👍

zerobias (approved):

Issue #680, opened 2022-05-14
RFC: create `useUnit` hook for React components


sergeysova:
Now we have two hooks: `useStore` and `useEvent`. But multiple usages of `useStore` will update the component twice or more, if some stores updated simultaneously. `useEvent` hook can bind not only events as well as effects.

Also, there is the `useEvent` proposal in the React ecosystem: https://github.com/reactjs/rfcs/pull/220

## Proposal

Create single hook `useUnit` accepts any kind of units and add batching under the hood.

```ts
const value = useUnit($store)
const handler = useUnit(someHappened)
const loadData = useUnit(loadDataFx)

// also supports mixing them all
const { value, handler, loadData } = useUnit({
  value: $store,
  handler: someHappened,
  loadData: loadDataFx,
})
```

I think `useUnit` should supports array shape:

```ts
const [handler, data] = useUnit([someHappened, $store])
```

## Use case

First of all it is simplification of the React bindings API. Using just one hook instead of two for the three kind of units.

Another use case is batching under the hood.

```ts
const { foo, bar } = useUnit({
  foo: $foo,
  bar: $bar,
})
```

At the moment, `useEvent` allows us to pass many events and effects:

```ts
const { handleA, handleB } = useEvent({
  handleA: someHappenedA,
  handleB: someHappenedB,
})
```

## Questions

**But what if someone passed two+ arguments?** Looks like we need to require the user to pass array, object, or just one argument explicitly. Otherwise there is will be too much function overloads. For example, there is will be two overloads with the same result:

```ts
const [first, second] = useUnit(a, b)
const [first, second] = useUnit([a, b])
```

## Another changes

Looks like `useStoreMap` and `useList` cannot be merged into `useUnit`. It is different kind of hooks.

AlexandrHoroshih:
What if objects with shape `{ store, fn, keys }` will be treated as a `useStoreMap`?

```ts
const { foo } = useUnit({
  foo: { store: $foo, fn: (f, [a,b]) => f ?? null, keys: [a, b] },
})
```

zerobias:
@AlexandrHoroshih it would be really cool if the types will allow that. Need to do a smoke test in the ts repl to prove that it will work in a first place

sergeysova:
@AlexandrHoroshih I suggest to split your proposal as a separate RFC

AlexandrHoroshih:
Released in `effector-react@22.1.0`

PR #685, opened 2022-05-17
Add SolidJS bindings

### LINKED ISSUE #658: Binding for `SolidJS`
## Proposal

Add new package `@effector/solid` with helpers (`useStore`, `useEvent`) for connecting Effector and [SolidJS](http://solidjs.com/).

## Use case

1. SolidJS is getting more popular.

Screenshot 2022-04-11 at 13 20 39

2. SolidJS is a really great way to build interfaces, but the internal reactivity system is pretty limited. So, we can improve DX for every developer who uses SolidJS.

P.S. This issue is for progress tracking, I'm going to write this bridge in a couple of weeks.

Drevoed:
### Conventions

Resolves #658


docs/api/effector-solid/useEvent.md
```diff
@@ -0,0 +1,99 @@
+---
+id: useEvent
+title: useEvent
+---
+
+Bind event to current fork instance to use in dom event handlers. Used **only** with ssr, in application without forks `useEvent` will do nothing
```
AlexandrHoroshih, line 6: I wouldn't agree, that `useEvent` is useful only with ssr 🤔 
e.g. it also can be useful if one wants to test components with something like RTL

Also i suggest to use `scope` instead of `fork instance`

I suggest something like this:
```suggestion
Bind event to current [_scope_](../effector/Scope.md) to use in dom event handlers.
Only `effector-solid/scope` version works this way, `useEvent` of `effector-solid` is no-op and does not require `Provider` with scope.
```

src/solid/lib/base.ts
```diff
@@ -0,0 +1,122 @@
...
+  });
+
+  onCleanup(stop);
+
+  return createMemo(() => result().val)
+}
+
+function updateRef<State, Keys, Result>(
+  sourceValue: State,
+  keys: Keys,
```
AlexandrHoroshih, line 98: Does it actually makes sense to have keys in implementation for solid? 🤔 

Looks like in this case there is no real optimization value

AlexandrHoroshih (approved):
Great work 👍

packages/effector-solid/package.json
```diff
@@ -0,0 +1,105 @@
+{
+  "name": "effector-solid",
+  "version": "1.0.0",
```
AlexandrHoroshih, line 3: Should we release it as `1.0.0`?

I suggest to set version, that would allow breaking changes after first feedback from users, since current bindings are more of a port of react bindings, than specifically designed for solid 🤔

docs/api/effector-solid/useEvent.md
Drevoed: This was just a copypaste from docs of effector-react, might I change the wording there as well?

packages/effector-solid/package.json
sergeysova: I suggest to set version complement to another bindings. Like `0.22.0` to upgrade after release to `22.0.0` or `23.0.0`

packages/effector-solid/index.d.ts
```diff
@@ -0,0 +1,65 @@
+import { Accessor, Component } from "solid-js";
+import {Store, Event, Effect, Domain} from 'effector'
```
sergeysova, line 2: Looks like here is not formatted code

src/solid/lib/base.ts
```diff
@@ -0,0 +1,122 @@
+import { createWatch, is, Scope, Store } from "effector";
+import { throwError } from "./throw";
+import {
+  Accessor,
+  createMemo,
+  createSignal,
+  onCleanup
+} from "solid-js";
+
+const stateReader = <T,>(store: Store<T>, scope?: Scope) =>
```
sergeysova, line 10:
```suggestion
const stateReader = <T>(store: Store<T>, scope?: Scope) =>
```

Do we really need `,` here?

src/solid/nossr.ts
```diff
@@ -0,0 +1,25 @@
...
+    store: Store<State>
+    keys: Keys
+    fn(state: State, keys: Keys): Result
+    updateFilter?: (update: Result, current: Result) => boolean
+  }
+    | Store<State>,
+  separateFn?: (state: State, keys: Keys) => Result,
+): Accessor<Result> {
+  return useStoreMapBase([configOrStore, separateFn])
+}
```
sergeysova, line 25: What about to remove `useEvent` and `useStore` in favor of `useUnit`?

https://github.com/effector/effector/issues/680

@zerobias what do you think? What if we can do not add methods in the new bindings.

> But we still need `useStoreMap`

tools/builder/packages.config.ts
```diff
@@ -53,6 +53,7 @@ const version = {
   effector: '22.3.0',
   'effector-react': '22.0.6',
   'effector-vue': '22.1.0',
+  'effector-solid': '1.0.0',
```
sergeysova, line 56:
```suggestion
  'effector-solid': '0.22.0',
```

packages/effector-solid/package.json
Drevoed: True, will be more predictable

src/solid/lib/gate.ts
```diff
@@ -0,0 +1,98 @@
...
+  GateComponent.set = set
+  GateComponent.displayName = `Gate:${fullName}`;
+
+  return GateComponent;
+}
+
+export function createGate<Props>(
+  name: string = 'gate',
+  defaultState: Props = {} as any,
+): Gate<Props> {
```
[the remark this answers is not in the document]
Drevoed: To be honest I was thinking about making Gate a bit different for solid.

There is rarely a need to check for all props.

I was thinking if something like this can be done:

```ts
import {createGate} from 'effector-solid'
import {render} from 'solid-js/web'

const Gate = createGate('gate with props')

const App = () => (
  <section>
    <Gate foo="bar" />
  </section>
)

Gate.state.foo.watch(foo => {
  console.log('current state of foo prop', foo)
})

// => current state of foo prop undefined

const unmount = render(() => <App />, document.getElementById('root'))
// => current state of foo prop bar
```

But this will probably have to involve proxies and will introduce undefined behaviour when used like this

```ts
// @ts-ignore
Gate.state.thisPropWillNeverExist.watch(() => {})
```

src/solid/lib/base.ts
Drevoed: Keys is used not only as an optimization detail, it is also used in selector function to receive part of store.

docs/api/effector-solid/index.md
```diff
@@ -0,0 +1,19 @@
+---
+id: effector-solid
+title: API Reference
+---
+
+## Reactive helpers
+
+- [useStore(store)](./useStore.md)
```
igorkamyshev, line 8: As I understand now, `effector-react` is moving to simple hook like `useUnit`. Shall we switch this package to the same approach before release?

I think, it'll be harder to do such moves after release.

igorkamyshev: https://github.com/effector/effector/pull/685#discussion_r875774633

I got it 🤣

src/solid/nossr.ts
AlexandrHoroshih: I thought a little more and now i think it is better to start with `useUnit` in solid right away 🙂 

It is just one api instead of two, supports built-in batching, and easily usable just like old apis

Since bindings for solid are not yet bothered by being backward-compatible, i think, it is better to favor simpler api

Also, it is still will be easily rewritable from react like this:

From:
```ts
const a = useStore($a)
const b = useEvent(b)
```
to:
```ts
const a = useUnit($a)
const b = useUnit(b)
```
or:
```ts
const {a, b} = useUnit({a: $a, b})
```

src/solid/lib/gate.ts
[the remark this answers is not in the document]
AlexandrHoroshih: I think, it actually can work, if gate state shape will be defined explicitly 🤔 
```ts
const Gate = createGate({
 foo: null,
 propB: "default"
})

const App = () => (
  <section>
    <Gate foo="bar" />
  </section>
)

// only props, that were initially setup at the Gate, will be available on state
Gate.state.foo.watch(foo => {
  console.log('current state of foo prop', foo)
})

// error immideatly
// not defined in default gate shape - not available at `state`
Gate.state.thisPropWillNeverExist.watch(() => {})
```

src/solid/nossr.ts
Drevoed: Done!

src/solid/lib/gate.ts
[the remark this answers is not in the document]
Drevoed: Hmm, this might work, yeah.

src/solid/lib/base.ts
```diff
@@ -0,0 +1,187 @@
...
+
+  const seq = [
+    step.compute({priority: 'effect'}),
+    step.compute({priority: 'sampler', batch: true}),
+    step.run({
+      fn: () => {
+        batch(() => {
+          for (const store of stores) {
+            const key = storeIdMap[store.graphite.id]
+            storeSetterMap[key](() => store.getState())
```
AlexandrHoroshih, line 57: I think, it is better to use `stateReader` here 🤔 
```suggestion
            storeSetterMap[key](() => stateReader(store, scope))
```
so it explicitly reads from scope, when its needed

AFAIK, `store.getState()` actually will read value from scope, if called in its context, so current code will also work correctly, but since it is not a documented public feature, i think it sould be better to read from scope explicitly

sergeysova (approved):

src/solid/lib/base.ts
```diff
@@ -0,0 +1,187 @@
...
+
+  const initial = isList ? [] : {}
+  const stores = []
+  const storeIdMap = {}
+  const storeSetterMap = {}
+
+  for (const [key, value] of entries) {
+    if (is.store(value)) {
+      if (stores.includes(value)) {
+        throwError(`useUnit store at key "${key}" is already exists in shape`)
```
AlexandrHoroshih, line 36: Do we actually need to runtime-ban such cases? 🤔 

I can see usecases like this:
```ts
const useEllipseParams = (stores) => {
 return useUnit({
   leftFocalPoint: stores.leftFocalPoint,
   rightFocalPoint: stores.rightFocalPoint,
   ...
 })
}
```
where `left...` and `right...` can be the same store, if we want to draw a circle:
```ts
const circleParams = useEllipseParams({
 leftFocalPoint: $point,
 rightFocalPoint: $point
})
```
I think that in cases, where shape of `useUnit` argument have some meaning by itself, there sometimes can be the same store in different keys and we should not block it 🤔 

I guess, it should be more like this instead:

```ts
      const hasStore = Boolean(storeIdMap[value.graphite.id])
      
      if (!hasStore) {
        storeIdMap[value.graphite.id] = key
        const [get, set] = createSignal(stateReader(value, scope))
        initial[key] = get
        storeSetterMap[key] = set
      } else {
        const originalKey = storeIdMap[value.graphite.id]
        
        initial[key] = initial[originalKey]
        storeSetterMap[key] = storeSetterMap[originalKey] 
      }
```

Drevoed: True, also makes sense to fix effector-react version as well then

Drevoed:
I also want to suggest changing naming of the functions that the binding provides. In solid, functions which are used in reactivity system have `create*` names.  The reasoning for that is to differentiate more properly that the component body really executes only once. Instead of `useUnit` - `createUnit` seems more appropriate. `useGate` is an exception though, as is `useContext` in Solid.

AlexandrHoroshih (approved):

PR #686, opened 2022-05-20
Allow pass shape of Stores [v-model]

ilajosmanov:
Added support for passing shape of object. It would be great if developer uses something like effector-forms for vue.

```
const form = createForm({ ... })
form.$fields // shape of stores
```

After this update each field can be modified store separately


AlexandrHoroshih (approved):

zerobias:
Amazing! Will be published in `effector-vue 22.2.0`

PR #688, opened 2022-05-26
`safe` mode for `scopeBind`

### LINKED ISSUE #666: No-throw mode for `scopeBind`
## Proposal

Add an option for `scopeBind` to supress its exception in case, when there is no active scope.

```ts
const maybeBinded = scopeBind(event, { safe: true })
```
In this case, if `scopeBind` called with `safe: true` and there is no scope, then `scopeBind` just returns original event or a function, which calls original event (just like when scope is found)

## Use case

`scopeBind` is meant to be used to bind event or effect to a scope to be called later and the most common use case is to build a bridge of effector events to other reactive source, like websocket, browser api, etc

It works quite well, when used like this:
```ts
const connectSocketFx = attach({
 source: $socket,
 effect(socket) {
   socket.on("message", scopeBind(messageReceived))
 }
})

sample({
 clock: appStarted,
 target: connectSocketFx
})

// later
const scope = fork()

await allSettled(appStarted, {scope})
```
This way it is possible to set scope once, when app is initialized and all related subscribtions will also work in correct scope, which is also useful for tests

But if app doesn't use scopes and only needs Fork API for tests, then `scopeBind` usage is problematic because it throws exception, when scope is not found

And to allow usage both with and without scope users are required to write something like this:
```ts
const safeBind = (event) => {
 try {
   return scopeBind(event)
 } catch(e) {
   return event
 }
}

// later
const connectSocketFx = attach({
 source: $socket,
 effect(socket) {
   socket.on("message", safeBind(messageReceived))
 }
})
```

Built-in option would allow to solve these cases without custom helpers

AlexandrHoroshih:
Closes #666

- Added a new option to `scopeBind` - `safe: boolean`

### DIFF
--- a/packages/effector/index.d.ts
+++ b/packages/effector/index.d.ts
@@ -2999,7 +2999,7 @@ export function serialize(
  * @param unit event to bind
  * @returns function which will trigger an event in a given scope
  */
-export function scopeBind<T>(unit: Event<T>, opts?: {scope: Scope}): (payload: T) => void
+export function scopeBind<T>(unit: Event<T>, opts?: {scope?: Scope; safe?: boolean}): (payload: T) => void
 /**
  * Bind effect to a scope to be called later.
  *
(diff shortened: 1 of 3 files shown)


sergeysova:
@AlexandrHoroshih Can you resolve conflicts?

AlexandrHoroshih:
@sergeysova done ✅

sergeysova (approved):
Looks awesome. I think it can be published in the next __minor__ release.

Isn't it?

PR #730, opened 2022-06-18
chore: change eslint config's format from YAML to JSON

ainursharaev:


igorkamyshev:
Could you describe a purpose of this PR please?

sergeysova:
Because YAML is overengineered (it parses `off` as `false` for example). I see that JSON for eslint is most popular option

ainursharaev:
I was having problems with the current configuration in Webstorm, and I have discussed this issue with @sergeysova, and we decided to change the format to JSON.

Issue #684, opened 2022-05-14
When trigger `Gate.open` manually, props don't passed to `Gate.state`


sergeysova:
**What is the current behavior:**

```ts
const Gate = createGate({defaultState: 'empty'})
const scope = fork()
await allSettled(Gate.open, {scope, params: 'data'})

scope.getState(Gate.state) === 'data' // FALSE, 'empty'
```

**What is the expected behavior:**

Expected to have `data` in `Gate.state`

ilajosmanov:
We should close this issue.
Solved [here](https://github.com/effector/effector/pull/731)

PR #731, opened 2022-06-19
#684 fixed bug when .open calls manually but state aren't changing

ilajosmanov:
Fixed bug with .open event.
Also fixed case when open/close event calls but the payload of these events is incorrect


src/vue/composition.h.ts
```diff
@@ -1,8 +1,8 @@
 import {Domain, Event, Store} from 'effector'
 
 type Gate<Props> = {
-  open: Event<Props>;
-  close: Event<Props>;
+  open: Event<Props | void>;
```
igorkamyshev, line 4: 🤔 I do not get it, why `open` can be called without props?

src/vue/createGate.ts
```diff
@@ -34,13 +34,21 @@ export function useGate<Props>(GateComponent: Gate<Props>, cb?: () => Props) {
   }
 
   onMounted(() => {
-    const raw = unwrapProxy(_)
-    GateComponent.open(deepCopy(raw))
+    if (typeof _ !== "undefined") {
```
igorkamyshev, line 37: Same here

src/vue/composition.h.ts
ilajosmanov: @igorkamyshev why not? Sometimes I can use gate without passing props

src/vue/createGate.ts
ilajosmanov: answer above

src/vue/composition.h.ts
igorkamyshev: As I understand particular Gate should always be called with props or without props.

So I suggest to use Props type here and pass Props = void, if you are not going to use props.

ilajosmanov: Thanks! Fixed that

igorkamyshev (approved):
Looks good to me 👍

zerobias (approved):
Cool, thanks! 👍

PR #733, opened 2022-06-23
Tests and types for `useUnit`

AlexandrHoroshih:
- Adds missing tests and typings for brand new `useUnit`


zerobias (approved):
Amazing! 🔥 Thanks!

PR #739, opened 2022-07-11
Upgrade `effector-react` readme to `useUnit`

AlexandrHoroshih:
`useStore` -> `useUnit`

### DIFF
--- a/packages/effector-react/README.md
+++ b/packages/effector-react/README.md
@@ -50,7 +50,6 @@ const Form = () => {
 
 [Try it](https://share.effector.dev/gjsgk6oh)
 
-[useStore](https://effector.dev/docs/api/effector-react/useStore) in docs
+[useUnit](https://effector.dev/docs/api/effector-react/useUnit) in docs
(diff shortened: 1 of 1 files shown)


packages/effector-react/README.md
```diff
@@ -19,29 +19,27 @@ yarn add effector effector-react
...
-})
+const $size = $text.map(text => text.length)
 
 const Form = () => {
-  const {text, size} = useStore($form)
+  const {text, size} = useUnit({
+    text: $text,
+    size: $size,
+  })
+  const [handleTextChange] = useUnit([inputText])
```
sergeysova, line 36: Why not just `useUnit(inputText)` ?

AlexandrHoroshih: I wanted to show that different kinds of overloads are supported 🤔

sergeysova: I think we have explanation in the docs, but readme is just and example, isn't it?

zerobias:
Thanks!

PR #744, opened 2022-07-22
Add custom serialized for Store

### LINKED ISSUE #756: Custom serialization for stores
## Proposal

Currently we can only disable serialization of store's value via `serialize: 'ignore'` setting, but sometimes we do want to transfer state from server to the client, even if this state is represented in non-json shape - e.g. `Date` or `Map`

Idea is to add custom serialization config to stores:

```ts
const $map = createStore(new Map<number, number>(), {
      serialize: {
        write: map => [...map.entries()],
        read: jsonMap => new Map(jsonMap),
      },
})
```

Which would allow to introduce custom serialization to SSR usecases:
1. We have a store with non-serializable value - e.g. Map
```ts
    const $map = createStore(new Map<number, number>(), {
      serialize: {
        write: map => [...map.entries()],
        read: jsonMap => new Map(jsonMap),
      },
    })
```

2. During SSR we are creating separate `Scope` for each request, as usual, then run the logic
```ts
const scope = fork()

await allSettled(appStarted, {scope, params: req })
```

3. After logic is settled, we need to extract stores data to stringify it to JSON to transfer to client.
At this moment `serialize.write` method of `$map` will be called

```ts
const serverValues = serialize(scope)

serverValues[$map.sid] // has number[][], which is json-representation of Map<number, number>

const jsonString = JSON.stringify(serverValues) // no errors, which would happen because of Map
```

4. At the client we want to create a client scope and fill it with values from the server.
At this moment `serialize.read` method of `$map` is called

```ts
const clientScope = fork({
 values: serverValues
})

clientScope.getState($map) // Map<number, number> with values from the server

```

**Caveats**
We have two ways to setup `values` in `fork` call:
- `Map<Store, Value> | [Store, Value][]`, which is used for tests and sometimes as a di-mechanism
- `Record<string, Value>`, which is used for server-calculated values in the shape of json
We need a way to decide, when we need to call `serialize.read` methods and when we don't need it

I think, in the first case we have a clear user intent to set a specific value to specific store:
```ts
const scope = fork({
 values: [
    [$map, new Map(...)]
 ]
})
```
so here we don't need to deal with serialization at all.

The second case is not considered a public api for manual values setup - only to provide json with server values:
```ts
fork({
  values: {
    "-dsg4gf": [[0, 1], [2, 3]]
  }
})
```
so here we clearly need to use `serialize.read`

AlexandrHoroshih:
Closes #756 

### Conventions
- [ ] Please check your messages [against the guidelines](https://cbea.ms/git-commit/) if not, perform [interactive rebase](https://thoughtbot.com/blog/git-interactive-rebase-squash-amend-rewriting-history)


zerobias (approved):
Amazing, thanks! 👍

PR #761, opened 2022-08-17
Add recommendation to use `sample` instead of `forward`

YanLobat:
image


sergeysova:
Awesome!
@YanLobat Thank you

PR #763, opened 2022-08-18
Allow to pass domain as an argument for `createEvent` and similar methods

### LINKED ISSUE #590: Rework `domain`-related APIs
## Proposal

Current `domain`-related APIs is a hard to parse by AST-awared plugins. It leads to some problems and races with other tools (e.g., https://github.com/effector/effector/issues/580, https://github.com/effector/effector/issues/482, https://github.com/effector/effector/issues/204). But, `effector-react` uses the different approach and doesn't have this issue.

```ts
const domain = createDomain()
const Gate = createGate({ domain })
```

I propose to add the new optional parameter to all creator-functions.

```ts
// old approach
const domain = createDomain()
const $domainStore = domain.createStore(null)

// new approach
const domain = createDomain()
const $domainStore = createStore(null, { domain });
```

In next major releases, we can deprecate `domain.create*`-calls.

What do you think, folks?

sergeysova:
- Flatten config of unit not inside initUnit
- Start implementation of the {domain} in create* methods
- Refactor naming in createDomain
- Call domain hooks if domain passed as config argument

Closes #590 

### Conventions


src/effector/createDomain.ts
```diff
@@ -15,17 +15,35 @@ import {getGraph, getParent} from './getter'
...
-  const node = createNode({family: {type: DOMAIN}, regional: true})
+  const config = flattenConfig({
+    or: maybeConfig,
+    and: typeof nameOrConfig === 'string' ? {name: nameOrConfig} : nameOrConfig,
+  }) as any
 
-  const result = {
+  assert(
+    !(config?.domain && config?.parent),
+    'Domain cannot be created with two domains as parent. Please, remove {domain} argument.',
```
AlexandrHoroshih, line 29: We cannot throw exception here since it would block usage of `effector-root` and `effector-logger`, since they are relying on imports substitution 🤔 

In the actual code of the app we have:
```ts
const myCoolDomain = createDomain()

const event = createEvent({ domain: myCoolDomain })
```

But after we set up `effector-logger` via babel-plugin, we basically have the equivalent of
```ts
const myCoolDomain = loggerDomain.createDomain()

const event = loggerDomain.createEvent({ domain: myCoolDomain })
```
even though actual code of the app is unchanged

In case of throw it just would not be possible to make `logger` work without rewriting app code

I suggest to choose "parent domain override" behavior:
```ts
const event = d1.createEvent({domain: d2})
```
where `d2` would be parent domain of `event` and not `d1` - it would allow to use `effector-logger` and `effector-root` in the same way it was before, without actual changes in the app code

With `logger` it would work because every domain in the app will also belong to the loggerDomain

AlexandrHoroshih (changes requested):

src/effector/createDomain.ts
sergeysova: Makes sense. Implemented in 0d619620c5a0d700961c96bd0dcc0ff0b2cd7e68

src/effector/__tests__/domain.test.ts
```diff
@@ -333,15 +333,19 @@ describe('pass domain into creator', () => {
...
     const indirect = createDomain()
+    indirect.onCreateStore(indirectFn)
 
-    expect(() => {
-      // @ts-ignore check for runtime error
-      direct.createDomain({domain: indirect})
-    }).toThrowErrorMatchingInlineSnapshot(
-      `"Domain cannot be created with two domains as parent. Please, remove {domain} argument."`,
-    )
+    // @ts-expect-error There is no types for case, just runtime check
```
AlexandrHoroshih, line 345: Probably the other way around 🤔 
`there is no runtime check for the case, only types`

AlexandrHoroshih (approved):
Awesome 👍

igorkamyshev (approved):
LGTM

zerobias (approved):
Cool, thanks 👍👍 This will go to next _minor_ release as this is not a breaking change and there is a need to publish some features earlier

Issue #753, opened 2022-08-02
Documentation is not contain explaining what effector is and how it use


vitonsky:
Yesterday my friends did mention an effector, i've ask they question "what is effector? why i may prefer it to other state managers? which ideas its have?" and they just can't explain me and told "just see at examples".

I've visit a site https://effector.dev/ and read it, but i didn't found any information about effector concept and workflow, only useless examples that explain nothing.

This is extremely bad and useless documentation site, it contains only API information.

Could you add articles about:
- what is the "effector" and about differences of other state managers
- whats key features of effector and why i have to want to use effector. What's benefits effector will give me?
- about principles of effector. How i have to build architecture of my project that use effector?
- what's effector don't do and what i should to avoid? What's popular mistakes of effector use you know and want to warn me about?

I need a more information from effector developers about effector, to i can refer to docs instead of research each problem and explain to other developers in my team.

Explore this sites to understand content that docs site must contain:
- https://reactjs.org
- https://redux.js.org
- https://logux.io

PR #774, opened 2022-09-03
Add motivation page

sergeysova:
### What needs to be covered in total

- prerequisites — what is the reason to use effector
- problems it solves
- benefits
- trade-offs
- why do we want you to use it

The last two points will be covered in the next PRs

Relates #753

### DIFF
--- a/docs/introduction/motivation.md
+++ b/docs/introduction/motivation.md
@@ -1,0 +1,27 @@
+---
+id: motivation
+title: Motivation
+---
+
+Most applications have the logic for which they are created.
+This logic itself can be very complex and based on reactive principles, we call it **business logic**.
+
+In order for the user to benefit from the logic described above it is necessary to create a user interface
+and implement the logic for it, we call it **UI-logic**.
+
+The easiest way to figure out what is what — is to remember what tasks your project manager comes to you with,
+or what the product tasks look like. There's no description of how buttons, dropdowns, and form fields should work.
+There is a description of how the user interacts with the application.
+The users don't care what technologies and frameworks are used under the hood.
+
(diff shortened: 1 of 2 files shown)


docs/introduction/motivation.md
```diff
@@ -0,0 +1,27 @@
...
+Most applications have the logic for which they are created.
+This logic itself can be very complex and based on reactive principles, we call it **business logic**.
+
+In order for the user to benefit from the logic described above it is necessary to create a user interface
+and implement the logic for it, we call it **UI-logic**.
+
+The easiest way to figure out what is what — is to remember what tasks your project manager comes to you with,
+or what the product tasks look like. There's no description of how buttons, dropdowns, and form fields should work.
+There is a description of how the user interacts with the application.
+The users don't care what technologies and frameworks are used under the hood.
```
AlexandrHoroshih, line 15: I would say that there can be description of how buttons or dropdowns should work in product tasks - but this description will also be high-level and closer to users perspective.

Certainly, there will be no implementation details like "dropdown list state update should be React's low-priority transition" - it is up to the developer to decide, which frameworks and api's to use to achieve required level of quality.

I suggest to reflect this here too, something like that:
```suggestion
The easiest way to figure out what is what — is to remember what tasks your project manager comes to you with,
or what the product tasks look like.
There's no description of how buttons, dropdowns, and form fields should work internally or which api's should be used.
There is only a description of how the user interacts with the application.
The users don't care what technologies and frameworks are used under the hood - it is up to the developers to decide, which tools to use to achieve required quality.
```

AlexandrHoroshih (approved):
Looks awesome! 🔥 

I have a little suggestion on product tasks example, but everything else is 👍

igorkamyshev (approved):
Love it!

PR #776, opened 2022-09-05
Add forceScope parameter to useUnit in react

### LINKED ISSUE #765: Isomorphic `useUnit` in `effector-react`
## Proposal

`useUnit` from `effector-react` tries to find `scope` in the context and automatically choose correct mode:
- `scope`-full — binds events/effects to scope, reads store's values from scope
- `scope`-less — returns events/effects as is, reads global store's values 

Add optional parameter `forceScope` to imitate behaviour of current `useUnit` from `effector-react/scope`, throws error in `scope`-less case.

## Use case

Merge two packages and simplify `scope`-full cases in applications — SSR and tests.

sagdeev:
### Conventions

Resolves #765

### DIFF
--- a/src/react/__tests__/base/useUnit.test.tsx
+++ b/src/react/__tests__/base/useUnit.test.tsx
@@ -870,66 +870,130 @@ describe('useUnit', () => {
     afterEach(() => {
       console.error = consoleError
     })
-    it('Should throw error if no scope', async () => {
-      const $a = createStore(42)
+    describe('forceScope options not set', () => {
+      it('without context', async () => {
+        const $a = createStore(42)
 
-      const View = () => {
-        const a = useUnit($a, {forceScope: true})
+        const View = () => {
+          const a = useUnit($a)
 
-        return <div>{a}</div>
-      }
+          return <div>{a}</div>
+        }
 
-      expect(() => render(<View />)).rejects.toThrow(
-        'No scope found, consider adding <Provider> to app root',
-      )
+        expect(() =>
+          render(<View />).toMatchInlineSnapshot(`
+            <div>
+              42
+            </div>
+          `),
+        )
+      })
+      test('with context', async () => {
+        const $value = createStore('value')
+
+        const Component = () => {
+          const value = useUnit($value)
+          return <div>{value}</div>
+        }
+
+        const scope = fork({values: [[$value, 'scoped value']]})
+
+        await render(
+          <React.StrictMode>
+            <Provider value={scope}>
+              <Component />
+            </Provider>
+          </React.StrictMode>,
+        )
+
+        expect(container.firstChild).toMatchInlineSnapshot(`
+            <div>
+              scoped value
+            </div>
+          `)
+      })
     })
-    test('Should get value from scope', async () => {
-      const $value = createStore('value')
+    describe('forceScope disabled', () => {
+      it('without context', async () => {
+        const $a = createStore(42)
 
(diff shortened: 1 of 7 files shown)


igorkamyshev (approved):
Sounds great to me 💙 

Folks, @effector/core what do you think?

src/react/__tests__/base/useUnit.test.tsx
```diff
@@ -861,4 +861,75 @@ describe('useUnit', () => {
...
+        <React.StrictMode>
+          <Provider value={scope}>
+            <Component />
+          </Provider>
+        </React.StrictMode>,
+      )
+
+      expect(container.firstChild).toMatchInlineSnapshot(`
+          <div>
+            value
```
AlexandrHoroshih, line 930: Hmm, looks like there is a bit of misunderstanding 🤔 

In the original issue https://github.com/effector/effector/issues/765 there is a requirement for an isomorphic behavior, which basically means `useUnit` without `forceScope` will try to get scope from context and if its not found -> work like there is no scope

So this test should be updated this way:
```suggestion
           scoped value
```

AlexandrHoroshih (changes requested):
Hi @sagdeev !
Thanks for contribution!

Everything looks good, but there is a bit of misunderstanding, probably:

In the original issue `useUnit` without `forceScope` must try to find the scope from context and if scope is not found - work without it

So overall there a few things missing:
1. Last test snapshot with `forceScope: false` shoud have a value from scope
2. Need to add another test with `forceScope: false` and without Provider
3. Need to add isomorphic behaviour

src/react/nossr.ts
```diff
@@ -19,8 +20,8 @@ export function useStore<State>(store: Store<State>): State {
   return useStoreBase(store)
 }
 
-export function useUnit(shape) {
-  return useUnitBase(shape)
+export function useUnit(shape, opts?: {forceScope?: boolean}) {
+  return useUnitBase(shape, opts?.forceScope ? getScope() : undefined)
```
AlexandrHoroshih, line 24: I suggest to add isomorphic behaviour here something like this:
```suggestion
  return useUnitBase(shape, getScope(opts?.forceScope))
```

src/react/ssr.ts
```diff
@@ -18,7 +18,7 @@ import {deprecate} from './deprecate'
 
 const ScopeContext = React.createContext(null as Scope | null)
 export const {Provider} = ScopeContext
-function getScope() {
+export function getScope() {
   const scope = React.useContext(ScopeContext)
   if (!scope)
```
AlexandrHoroshih, line 23: It looks like a simplest place to introduce isomorphic behavoir to me, but maybe there a better ones 🤔 

I think, it may look like this:
```suggestion
export function getScope(forceScope?: boolean) {
  const scope = React.useContext(ScopeContext)
  if (forceScope && !scope)
```

sergeysova: I think we need to move `getScope` out from the `ssr.ts` file.
Because all export from these modules is available to the end-user.

sagdeev:
@AlexandrHoroshih hi! I think I understand what you were talking about 😃 I made some changes as you requested. Can you please take a look?

- now it works as expected
- update tests
- move `getScope` to new file and add exports

AlexandrHoroshih (approved):
@sagdeev Awesome, looks perfect 🔥👍

sergeysova (approved):

PR #780, opened 2022-09-08
added useStoreMap hook for vue 3 composition API

ilajosmanov:
Possibility to use `useStoreMap`


AlexandrHoroshih (approved):
Looks pretty cool!

However, shouldn't Vue users also be able to set a custom `updateFilter`?

ilajosmanov:
> Looks pretty cool!
> 
> However, shouldn't Vue users also be able to set a custom `updateFilter` like in React version?

That is a good question. We have two scenarios:
- passing an object
- passing primitive/array

If we're talking about the object - I patch this object for avoiding unnecessary updates. In this case, we have to think about how to fast compare objects

zerobias (approved):

PR #782, opened 2022-09-11
Add forceScope opts to useUnit in solidjs

### LINKED ISSUE #766: Isomorphic `useUnit` in `effector-solid`
Same as https://github.com/effector/effector/issues/765

usmanyunusov:
### Conventions

Fix #766

### DIFF
--- a/src/solid/scope.ts
+++ b/src/solid/scope.ts
@@ -1,25 +1,13 @@
-import {
-  Accessor,
-  createContext,
-  createEffect,
-  createMemo,
-  useContext,
-} from 'solid-js'
-import {Domain, Scope} from 'effector'
-import {throwError} from './lib/throw'
-import {useStoreMapBase, useUnitBase} from './lib/base'
-import {useGate as commonUseGate, createGateImplementation} from './lib/gate'
+import {Accessor, createEffect, createMemo} from 'solid-js'
+import {Domain} from 'effector'
+
 import type {Gate} from './index.h'
 
-const ScopeContext = createContext<Scope | null>(null)
-export const {Provider} = ScopeContext
+import {useGate as commonUseGate, createGateImplementation} from './lib/gate'
+import {useStoreMapBase, useUnitBase} from './lib/base'
+import {getScope, ScopeContext} from './lib/get-scope'
 
-function getScope() {
-  const scope = useContext(ScopeContext)
-  if (!scope)
-    throwError('No scope found, consider adding <Provider> to app root')
-  return scope as Scope
-}
+export const {Provider} = ScopeContext
 
 export function createGate<Props>(
   config: {
@@ -59,7 +47,8 @@ export function useGate<Props>(
   createEffect(() => commonUseGate(ForkedGate(), props))
 }
 export function useUnit(shape) {
-  return useUnitBase(shape, getScope())
+  const scope = getScope()
+  return useUnitBase(shape, scope)
 }
 /** useStoreMap wrapper for scopes */
 export function useStoreMap(configOrStore: any, separateFn: any) {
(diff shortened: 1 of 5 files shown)


src/solid/scope.ts
```diff
@@ -11,12 +11,12 @@ import {useStoreMapBase, useUnitBase} from './lib/base'
 import {useGate as commonUseGate, createGateImplementation} from './lib/gate'
 import type {Gate} from './index.h'
 
-const ScopeContext = createContext<Scope | null>(null)
+const ScopeContext = createContext(null as Scope | null)
 export const {Provider} = ScopeContext
 
-function getScope() {
+export function getScope(forceScope?: boolean) {
```
igorkamyshev, line 17: In my opinion, it is better not to export this function from public module, let's move it somewhere inside the package?

igorkamyshev (commented):
Thanks for a contribution 💙

src/solid/scope.ts
usmanyunusov: Maybe take move everything the file `scope.ts` into an `ssr.ts` file. Leave only public methods in the `ssr.ts`.

src/solid/scope.ts
```diff
@@ -58,8 +58,9 @@ export function useGate<Props>(
 
   createEffect(() => commonUseGate(ForkedGate(), props))
 }
-export function useUnit(shape) {
-  return useUnitBase(shape, getScope())
+export function useUnit(shape, opts?: {forceScope?: boolean}) {
+  const scope = getScope(opts?.forceScope)
```
AlexandrHoroshih, line 62: I think that `/scope` version of bindings should have `forceScope` enabled by default 🤔 

I suggest to change it like that:
```suggestion
  const scope = getScope(opts?.forceScope ?? true)
```

src/solid/scope.ts
AlexandrHoroshih: Yep, that would be better move 🤔 

`getScope` should not be public, that's for sure

AlexandrHoroshih (changes requested):
Thank you for your contribution! 🧡 

Everything looks good, all test in place, feature works as expected 🔥  👍 

There only few small caveats:

src/solid/scope.ts
AlexandrHoroshih: Hmm, i thougth a little more and now i suggest to do it like this:


1. Inside `getScope` there should be default fallback like `forceScope = true`

2. For `/scope` exports we don't need to do anything besides `getScope()` call without params

3. For main `effector-solid` exports we should do something like `getScope(ops?.forceScope ?? false)` inside

usmanyunusov:
I don't pass the `scopes.test.tsx` solid test . What is the right way to test?

AlexandrHoroshih:
Hmm, @Drevoed maybe you can help? 🤔 

I don't quite understand, what is happening in the tests and why changes in PR would affect this

AlexandrHoroshih:
@usmanyunusov nevermind, i got it.

You moved hooks into `ssr.ts`, while they were placed at `scope.ts` - it has broken the build

I suggest to revert these changes (get hooks back into `scope.ts`) and move `getScope` and related Scope context into file named like `scope-context`

AlexandrHoroshih:
`ssr.ts` would not work with Solid.js, because it never had such sub-export

There is one in React bindings, but it is a leftover of past times, when there were no other usecases for fork-api, than ssr

Now it is deprecated, basically, as `/scope` is much better naming now

In solid.js bindings it was done this way from the start

usmanyunusov:
@AlexandrHoroshih, Fixed

AlexandrHoroshih (approved):
Looks good!

@Drevoed what you think?

Drevoed:
LGTM, awesome work! Thanks :) :rocket:

PR #801, opened 2022-10-09
Protect against combine argument being broken via Array.slice

AlexandrHoroshih:
Effector uses Array.slice internally to prepare argument object for a user, there is a case, where it was broken

Changelog:
- Added tests against breaking `combine` argument in userland via `.slice = undefined` or similiar code
- Changed `arr.slice()` to `[...arr]`


zerobias (approved):
Great! Thanks 👍

PR #807, opened 2022-10-22
Support `fork(scope)` edge cases

AlexandrHoroshih:
There is a potential breaking change at the types level - now all settled will return special `{ status: 'forked' }` object, if scope was forked during allSettled run

Typescript will force user to explicitly handle this new case

### DIFF
--- a/packages/effector/index.d.ts
+++ b/packages/effector/index.d.ts
@@ -3095,7 +3095,7 @@ export function fork(config?: {
 export function allSettled<FX extends Effect<any, any, any>>(
   unit: FX,
   config: {scope: Scope; params: EffectParams<FX>},
-): Promise<{status: 'done', value: EffectResult<FX>} | {status: 'fail'; value: EffectError<FX>}>
+): Promise<{status: 'done', value: EffectResult<FX>} | {status: 'fail'; value: EffectError<FX>} | {status: 'forked'}>
 /**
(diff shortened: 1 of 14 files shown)


src/effector/fork/createScope.ts
```diff
@@ -127,7 +128,31 @@ export function createScope(unit?: Domain): Scope {
...
+    // Update old scope scopeRef
+    // to redirect all effect.finally and scope-binded events to new scope
+    oldScope.scopeRef.ref = resultScope
+
+    // transfer old scope scopeRef to the new scope,
+    // so any further fork(scope) calls will also redirect any effect.finally and scope-binded events to new scope
+    resultScope.scopeRef = oldScope.scopeRef
+
+    // transfer old scope state to the new one
+    resultScope.sidValuesMap = Object.assign({}, oldScope.sidValuesMap)
```
zerobias, line 147: Usually we use object spread except cases with mutable assignments
```suggestion
    resultScope.sidValuesMap = {...oldScope.sidValuesMap
```

packages/effector/index.d.ts
```diff
@@ -3001,6 +3001,7 @@ export function combine<T extends Tuple<Store<any>>>(
  * Fully isolated instance of application. The primary purpose of scope includes SSR and testing
  */
 export interface Scope extends Unit<any> {
+  live: boolean;
```
sergeysova, line 3004:
```suggestion
  alive: boolean;
```

PR #812, opened 2022-11-09
added express24 logo

45m0d3u5:


hexagon141:
also leaved comment in #278

Issue #813, opened 2022-11-21
Fork handlers effect error


ivanov-v:
**What is the current behavior:**
Typescript generated error if effect contains custom error `effect<void, void, CustomError>`

**Please provide the steps to reproduce and if possible a minimal demo of the problem via https://share.effector.dev, https://codesandbox.io or similar**

https://www.typescriptlang.org/play?ts=4.7.4#code/JYWwDg9gTgLgBAbzgYygUwIYzQUQGZ5rIwA0ce0A1nAL7lQQhwDkaBRM0zA3AFC-IIAOwDO8NoWIAxAB5wAvCnRZc7YgB4AbhGAATMtr1kEINCJEYA5mgBccMVGBDLNAHwAKdwEoFrxDS8+AWExe0EwNAVyKncEXjg4AAsMIV0AGzQoETsAbXiEuByJDlkyb19-AF0SfOreAN4gA

**What is the expected behavior:**
No error

**Which versions of effector packages, and which browser and OS are affected by this issue? Did this work in previous versions of effector?:**
Effector: 22.4.0
Typescript:  > = 4.7.4

PR #814, opened 2022-11-21
Allow effects with custom error type in fork handlers

ivanov-v:
Closes: https://github.com/effector/effector/issues/813


AlexandrHoroshih (approved):
Awesome! Thank you 🔥

PR #822, opened 2022-12-15
`@@unitShape` protocol

igorkamyshev:
- [x] handbook about protocol for ecosystem's developers

### DIFF
--- a/docs/ecosystem-development/unit-shape-protocol.md
+++ b/docs/ecosystem-development/unit-shape-protocol.md
@@ -58,7 +58,7 @@ function createRoute(/* ... */) {
 
 Q: How frequently `@@unitShape`-function is called?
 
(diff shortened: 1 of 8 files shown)


docs/ecosystem-development/unit-shape-protocol.md
```diff
@@ -0,0 +1,67 @@
...
+}
+```
+
+## FAQ
+
+---
+
+Q: How frequently `@@unitShape`-function is called?
+
+A: Normally, it depends on a UI-library. For example, `effector-react` calls it once per component render, but `effector-solid` calls it once per component mount.
```
AlexandrHoroshih, line 61:
```suggestion
A: As many times as `useUnit` itself is called - it depends on a UI-library. For example, `effector-react` calls it once per component render, but `effector-solid` calls it once per component mount.
```

AlexandrHoroshih (approved):
🔥

zerobias (approved):

PR #837, opened 2023-01-14
Forbid restore(event) and restore(effect) in types

yumauri:
Calling `restore` with Event or Effect without second `defaultState` argument will fail with runtime error.
But typescript will not warn about it, so you can miss this until you actually run the code.
I tried to add some type guards against it.


zerobias (approved):
Great, thanks!

Issue #755, opened 2022-08-11
effector Spacewatch 23 tracking issue


sergeysova:
- [x] Remove flow typings from this repo #912
- [x] Remove deprecated options and methods
	- [x] Call of derived event #905
	- [x] Unit call from pure function #905
	- [x] #846
	- [x] #843 
	- [x] #842
	- [x] #845
	- [x] #844
	- [x] Remove `createStoreObject` #908
	- [x] #1018
- [x] Introduce new types
	- [x] Introduce `EventCallable<T>` and `StoreWritable<T>`. #847 (#966) 
	- [x] Forbid derived units in operators `target`s
- [x] Deprecate some features
	- [x] `forward` and `guard` #913
	- [x] Rename `greedy` into `batch` in `sample` (#972)
		- [x] Deprecate `greedy: false`, introduce `batch: true`
	- [x] #848
   - [x] Remove fork(scope)-related code (activeEffects, etc)
	- [x] Deprecate `reactSsr` option in babel-plugin (#910)
	- [ ] Deprecate `bindings` option in swc plugin (@Drevoed) https://github.com/effector/swc-plugin/#bindings
	- [x] Deprecate onlyChanges option in serialize #907
	- [x] Deprecate `{[fx.sid]: Function}` in fork handlers (#973)
	- [x] Deprecate `.watch(trigger, watcher)` #906
	- [x] Deprecate `getType()` (#899)
	- [x] Deprecate `effector/babel-plugin-react` #948
	- [x] Deprecate `fork(domain)` (https://github.com/effector/effector/pull/950)
	- [x] Deprecate `hydrate(domain)` (https://github.com/effector/effector/pull/950)
	- [x] Deprecate `hydrate(fork(domain))` — hydrating scope forked from domain (https://github.com/effector/effector/pull/950)
- [x] #903
- [x] https://github.com/effector/effector/issues/633 (#849)
- [x] https://github.com/effector/effector/issues/535 (#909)
- [x] Unify programmatic sid insertions: `sample({and: arg, or: {sid, name})` vs `restore(arg, arg, {sid, name})` (#969)
- [x] Throw an exception if `fn` in `$store.map(fn)` returns `undefined` in the initialization
      (#828)
- [x] Replace `createWatch` from the internal `effector-react` impl with `createWatch` from the `effector` package (related: https://github.com/effector/effector/pull/646) (#850)
- [x] Add `batch` option to `createWatch` (#850)
[...]

PR #842, opened 2023-02-02
Remove effector/fork module

zerobias:
Related issue: #755


zerobias:
At this moment I'm only updating the runtime and types, it would be great if you update this pr with docs improvements 🙏

Issue #784, opened 2022-09-12
ts error when using combine with primitive values 


vitalik-l:
Example from the docs doesn't work with typescript (https://effector.dev/docs/api/effector/combine/#combine-with-primitives-and-objects)

```ts
const $a = createStore('a')
const b = 2
const c = [false]
const d = { value: 1 }

const $resultUsingComa = combine($a, b, c, d)
```

TS error:
```
No overload matches this call.
  Overload 1 of 18, '(a: Store<string>, b: Store<unknown>, c: Store<unknown>, fn: (a: string, b: unknown, c: unknown) => unknown): Store<unknown>', gave the following error.
    Argument of type 'number' is not assignable to parameter of type 'Store<unknown>'.
  Overload 2 of 18, '(stores_0?: Store<any> | undefined, ...stores_1: Store<any>[]): Store<[(Store<any> | undefined)?, ...any[]]>', gave the following error.
    Argument of type 'number' is not assignable to parameter of type 'Store<any>'.
```

## Relevant documentation
https://effector.dev/docs/api/effector/combine/#combine-with-primitives-and-objects

PR #884, opened 2023-04-02
fix(784) combine typings, allow to pass primitives

faustienf:
issue https://github.com/effector/effector/issues/784

`combine($store, 1, '2')` should allow passing primitives, not only stores ([docs](https://effector.dev/docs/api/effector/combine/#combine-with-primitives-and-objects))

image

image


faustienf:
I've found that the combine() tests don't have cases with primitives. I'll add a test case too.

PR #891, opened 2023-04-21
beta: Fully rewrite `Event` API page

sergeysova:
[Rendered API/Event](https://github.com/effector/effector/blob/docs/event-rewrite/beta/src/content/docs/en/api/effector/Event.md)

[Rendered Explanation/Events](https://github.com/effector/effector/blob/docs/event-rewrite/beta/src/content/docs/en/explanation/events.md)

### DIFF
--- a/beta/src/content/docs/en/api/effector/Event.md
+++ b/beta/src/content/docs/en/api/effector/Event.md
@@ -11,7 +11,7 @@ redirectFrom:
 
 The **Event** in effector represents a user action, a step in the application process, a command to execute, or an intention to make modifications, among other things. This unit is designed to be a carrier of information/intention/state within the application, not the holder of a state.
 
-There are two type of events provided by effector: [`Event`](#event) and [`ListenEvent`](#listenEvent).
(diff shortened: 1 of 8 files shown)


beta/src/content/docs/en/api/effector/Event.md
```diff
@@ -9,131 +9,272 @@ redirectFrom:
...
 
-```js
+# Event instance {#event}
+
+In most situations, it is recommended to create events directly within the module, rather than placing them within conditional statements or classes, in order to maintain simplicity and readability. An exception to this recommendation is the use of factory functions; however, these should also be invoked at the root level of the module.
+
+:::info
+Event instances persist throughout the entire runtime of the application and inherently represent a portion of the business logic.
+
+Attempting to delete instances and clear memory for the purpose of saving resources is not advised, as it may adversely impact the functionality and performance of the application.
```
AlexandrHoroshih, line 23: Probably this part is not really needed here - it just adds to much of additional complexity to explanation. I think, it should be explained somewhere else, as a separate article

PR #895, opened 2023-05-11
Allow `domain` to be used in `attach` without parent effect

### LINKED ISSUE #829: Allow to pass `domain` to `attach` without parent effect
It is useful only when `attach` doesn't have a parent domain.

## Proposal

Propose to add `domain` for the `attach` operator when parameters `source` and `effect` appeared.

```ts
const domain = createDomain()
const $store = createStore(0)

const fx = attach({
  domain,
  source: $store,
  async effect(source, params: void) {
    // logic
  },
})
```

## Use case

We need to cover cases when created effect needs to be handled by the effect's hook `.onCreateEffect`, but in previous versions, there was no ability to do this.

Related #763

kireevmp:
Closes #829
Related #763

### DIFF
--- a/docs/api/effector/attach.md
+++ b/docs/api/effector/attach.md
@@ -161,6 +161,7 @@ Create effect which will call async function with values from `source` stores
 let resultFx: Effect<Params, Result>
 resultFx = attach({
   source: Store<Source>,
+  domain?: Domain,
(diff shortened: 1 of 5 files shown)


sergeysova (approved):
LGTM

zerobias (approved):

PR #902, opened 2023-05-17
Fix type degradation from 22.8.2 

igorkamyshev:
Some degradation of typings was introduced in https://github.com/effector/effector/pull/884/files


zerobias (approved):
Great, thanks!

PR #907, opened 2023-05-20
Deprecate `onlyChanges`

zerobias:


darky:
Why it was deprecated? How to serialize `onlyChanges` in future Effector 24?

kireevmp:
Hey @darky,  
Starting from Effector 22, `onlyChanges: true` is the default option for `serialize` ([changelog](https://changelog.effector.dev/#effector-22-0-0)).

In this PR, we're only deprecating the old `fork` + `serialize` behavior, which specifically required explicitly passing a Domain to track units. 

Since v22, Domain is no longer necessary, so both `fork(domain)` and `serialize(scope, { onlyChanges: false })` are deprecated.

darky:
@kireevmp thanks for explanation! 👍

PR #909, opened 2023-05-24
Fix global state impact on scopes

### LINKED ISSUE #535: No scope events affect scoped stores
Out of scope event calling affects store value within the scope. But it stops affecting right after first store value extraction from scope

https://share.effector.dev/4dHN6bJ8

Log from REPL
```
1.1 2 2
1.2 4 2
2.1 0 0
2.2 4 0
```

It shouldn't affects

Expected log from REPL
```
1.1 2 0
1.2 4 0
2.1 0 0
2.2 4 0
```

reproduce at 22.1.2, don't reproduce at 21.8.12

zerobias:
fix #535


AlexandrHoroshih (approved):
🔥

PR #927, opened 2023-06-30
EFX-167: Add effector/compat docs to beta docs

igorkamyshev:
Closes EFX-167


beta/src/content/docs/en/api/effector/moduleCompat.md
```diff
@@ -0,0 +1,86 @@
...
+      effector: "effector/compat",
+    },
+  },
+});
+```
+
+</details>
+
+:::warning{title="Bundler, not transpiler"}
+Since third-party libraries can import `effector` directly, you should not use transpilers like Babel to replace `effector` with `effector/compat` in your code, because by default Babel will not transform third-party code. Use bundler instead, it will replace `effector` with `effector/compat` in all modules, including third-party ones.
```
AlexandrHoroshih, line 85:
```suggestion
Since third-party libraries can import `effector` directly, you **should not** use transpilers like Babel to replace `effector` with `effector/compat` in your code, because by default Babel will not transform third-party code.

**Use bundler instead**, as it will replace `effector` with `effector/compat` in all modules, including third-party ones.
```

beta/src/content/docs/en/api/effector/moduleCompat.md
```diff
@@ -0,0 +1,86 @@
+---
+title: effector/compat
```
AlexandrHoroshih, line 2: I think, `description` and `redirectFrom` settings should be also added here, as for other pages in the docs 🤔

AlexandrHoroshih (approved):
Awesome!

beta/src/content/docs/en/api/effector/moduleCompat.md
igorkamyshev: Description is added ✅ 

But for `redirectFrom` I am not certain. In old docs, this module was described in the description of the installation process — https://effector.dev/docs/introduction/installation/#compatibility. So, we do not have a separate page for it to redirect from.

beta/src/content/docs/en/api/effector/index.md
```diff
@@ -55,6 +55,12 @@ List of the methods
 - [launch](/en/api/effector/launch)
 - [inspect](/en/api/effector/inspect)
 
+## Import map
+
+Package `effector` provides couple different entry points for different purposes:
+
+- [effector/compat](/en/api/effector/moduleCompat)
```
sergeysova, line 62: What about just:
```suggestion
- [effector/compat](/en/api/effector/module/compat)
```

beta/src/content/docs/en/api/effector/moduleCompat.md
```diff
@@ -0,0 +1,89 @@
...
+  resolve: {
+    alias: {
+      effector: "effector/compat",
+    },
+  },
+});
+```
+
+</details>
+
```
sergeysova, line 84: What about adding a babel guide with [`babel-plugin-module-resolver`](https://www.npmjs.com/package/babel-plugin-module-resolver)?

```json5
// .babelrc
{
  "plugins": [
    ["module-resolver", {
      "root": ["./src"],
      "alias": {
        "effector": "effector/compat",
      }
    }]
  ]
}
```

sergeysova (changes requested):

PR #960, opened 2023-08-14
Fix layout on mobile devices

MiiZZo:
### Changes
A little fix in companies section layout on mobile devices

### Before
before
### After
after


sergeysova (approved):
Thank you

PR #981, opened 2023-09-24
Forbid derived units in target

AlexandrHoroshih:


packages/effector/index.d.ts
```diff
@@ -155,6 +156,7 @@ export interface Event<Payload> extends Unit<Payload> {
  * The function you can call to trigger an event.
  */
 export interface EventCallable<Payload> extends Event<Payload>, UnitTargetable<Payload> {
+  kind: "event"
```
Minhir, line 159: Why we don't go futher and introduce `kind: "event-callable"`?

AlexandrHoroshih: Why do we need to?

So far it is not really necessary, i think - this move was done only to improve TS type inference in a few edge-cases

AlexandrHoroshih: Also there is now a separate runtime field `unit.targetable` 😅

Minhir: I thought more about consistency. Never mind, it looks good for me.

Minhir (approved):

src/effector/is.ts
```diff
@@ -54,9 +54,8 @@ export const assertTarget = (
   targetField: string = 'target',
 ) =>
   forEach(arrifyNodes(target), item =>
-    deprecate(
+    assert(
       !getMeta(item, 'derived'),
-      `${method}: derived unit in "${targetField}"`,
-      `createEvent/createStore`,
+      `${method}: derived unit in "${targetField} is not supported, use createStore/createEvent instead"`,
```
zerobias, line 59: Quote is missing
```suggestion
      `${method}: derived unit in "${targetField}" is not supported, use createStore/createEvent instead"`,
```

PR #1184, opened 2024-11-03
[Translate RU] Update navigation

Olovyannikov:
- Adds translation to navigation for sections that already have translation;


zerobias (approved):
Thanks!

PR #1212, opened 2025-01-04
feat(docs): add page about scope in advanced topic

sentend:
### Conventions


documentation/src/content/docs/en/advanced/work-with-scope.mdx
```diff
@@ -0,0 +1,501 @@
...
+  }
+}
+
+async function asyncProcess() {
+  console.log("here is ok", value); // effector
+
+  await 1;
+
+  // here we already lost context
+  console.log("but here is not ok ", value); // undefined
```
zerobias, line 469: `value` is not changed, it's better to use an example from [Scope](https://effector.dev/en/api/effector/scope/#scope-loss) article, as it guaranteed to be valid and concise

sentend: why is the value not changing? I am changing the value to `458`
i took the information from the current Scope page as a basis, and added the reason why the loss occurs, which is taken from proposal

> `value` is not changed, it's better to use an example from [Scope](https://effector.dev/en/api/effector/scope/#scope-loss) article, as it guaranteed to be valid and concise

zerobias: Oh, I see now. Anyway its better to show issue without additional indirection, just `scope.getState`

zerobias (approved):
Cool!

PR #1216, opened 2025-01-14
feat(docs): add info about skipVoid in manage-states

sentend:
### Conventions

### DIFF
--- a/documentation/src/content/docs/en/essentials/manage-states.mdx
+++ b/documentation/src/content/docs/en/essentials/manage-states.mdx
@@ -439,19 +439,36 @@ const $form = createStore({ email: "", password: "" })
   .reset([formSubmitted, formReset]);
 ```
 
-## `undefined` and `null` values (#void-values)
+## `undefined` Values (#void-values)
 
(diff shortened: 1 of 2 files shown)


documentation/src/content/docs/en/essentials/manage-states.mdx
```diff
@@ -439,4 +439,46 @@ const $form = createStore({ email: "", password: "" })
   .reset([formSubmitted, formReset]);
 ```
 
+## `undefined` and `null` values (#void-values)
+
+By default, effector skips updates with `undefined` or `null` values. This is done to prevent accidental errors. If you need to use `undefined` as a valid value, you need to explicitly specify `skipVoid: false` when creating a store :
```
zerobias, line 444: Only `undefined` is skipped, it's very important as `null` is indeed a replacement for undefined. We will disable this behavior (thats is what whole `skipVoid` story is about), this need to be mentioned too

The goal was to be able to not return anything from reducers when is update is not needed. Turns out that is better to just return old value

zerobias (changes requested):

zerobias (approved):
Great, thanks!

Issue #999, opened 2023-10-22
`split` cases show error with inline prepend calls


zerobias:
When event created by prepend directly in cases object, it shows errors in any cases, but everything works ok when unit or cases object created in separated variable

This not works:

```ts
split({
  source: fooFx.failData,
  match: (res) => 'ok',
  cases: {
    __: errorSettled.prepend(() => 'ok')
  }
})
```

This works:

```ts
const resultTarget = errorSettled.prepend(() => 'ok')

split({
  source: fooFx.failData,
  match: (res) => 'ok',
  cases: {
    __: resultTarget
  }
})
```

```ts
const cases = {
  __: errorSettled.prepend(() => 'ok')
}

split({
  source: fooFx.failData,
  match: (res) => 'ok',
  cases
})
```

[Link to TS repl](https://www.typescriptlang.org/play?ts=5.2.2&ssl=32&ssc=3&pln=14&pc=1#code/JYWwDg9gTgLgBAbwMZQKYEMaoKIDNepIwA0cAzmADbAlwoZbYBuqAdrc2zAL5y5QQQcAOSp8hGNGEAoaUgisy8XBAgAxAB5wAvHTSYc4ogApjASh0A+RNzNyFSuKigCoAZVQwYlVABMdegw4LOwAPEpQwKwA5pbmshTUMMYI0nDkEACuUEioAFx8qpoAdLjowJQAIpjoxGlwIJhIABYFxmhkFtrWwhAA1sJ16UjoZKhkBanp6QD6MwXOrh5ePr7FYGhgbL6mXT39wnbp3NK2svKK8B2ZlDAAKuhQ0Z4Bi9DL3n7rm9u7ViIHOzSRI0FL1MhZHL5QrqDSlcpVGpDBpNVpwdrjPYAgbIkZjCaIeqzeZwa63B5PTz1E5neyXOijcYBKZwOYLFzvTyfNYbVBbVg7cz-XoDOwnYFUUEsiHZXIFFSw+EVaowWr1RowFptDpYkWDep48anMxAA)

L117:
This is a TypeScript type inference nuance, `cases` in this case (No puns intended) is something like `Record<string, Event<T>>`. This issue can be easily solved by adding `as const` like this:
```ts
split({
  source: fooFx.failData,
  match: (res) => 'ok',
  cases: {
    __: errorSettled.prepend(() => 'ok')
  } as const
});
```

Issue #769, opened 2022-08-21
Typeguards in `match` of `split` are not working in config overload


AlexandrHoroshih:
**What is the current behavior:**
Currently typeguards in `match` field are not working in split

Following code
```ts
const $s = createStore<string | null>(null)
const valid = createEvent<string>()

const clk = createEvent()

split({
    source: $s,
    clock: clk,
    match: {
        valid: (str: string | null): str is string => Boolean(str),
    },
    cases: {
        valid,
    }
})
```
gives type error like `{ error: "source type should extends cases"; sourceType: string | null; caseType: string; }`

It can be workarounded by using other overload, where `split` actually creates events based on `match` typings
```
const $s = createStore<string | null>(null);
const valid = createEvent<string>();

const clk = createEvent();

const { validCase } = split(
  sample({ source: $s, clock: clk }),
  {
    validCase: (str: string | null): str is string =>
      Boolean(str),
  }
);

sample({
  clock: validCase,
  target: valid,
});
```

**Please provide the steps to reproduce and if possible a minimal demo of the problem via https://share.effector.dev, https://codesandbox.io or similar**
- https://tsplay.dev/wORPyW - broken example
- https://tsplay.dev/mLyleW - workaround

**What is the expected behavior:**
- Typeguards in `match` of config form of `split` are working
**Which versions of effector packages, and which browser and OS are affected by this issue? Did this work in previous versions of effector?:**
current latest release of `effector@22`

zerobias:
We are found a [novel approach](https://github.com/effector/effector/pull/1139) to improve sample types, it very similar to split in many ways, so after work will be done, it will be possible to port these improvements to split, thereby fixing this and many other issues in split types. I can’t promise to do this in 24.0.0 release so help will be appreciated

Issue #622, opened 2022-02-02
split + attach error in 22.2.0


uulaal:
[Playground](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgYygUwIYzQUQGZ5rIwA0cAzmADbCkrpZoDKM0aZWMGyAFnAL5w8UCCDgByNASKso4gNwBYAFArkEAHbl4AEg0BXEACM0UOAF56mbCzYAeA8dMA+ABQAGAJRLVyyjRhXBBU4UIoIfShkNAAuOD1DEygSELCQLF44hH4U5TCUDHI0cizU-LgAfQq4zm4eILLy-PIIqNirRlt0OwwNAE83AyoqT1ympqlCYjjXCrIwDCgMEDjHJM8LZ0R+RrD+Dcad5X2VFX9aBrywlsjouISnZMb0mEztsbDkQuLSq-KquKoay4aTEByJUxkABuEGAABM3BsAPRIuCAdhBACwggFYQQCCIIAZEDggCYQHGARRA4IA+EEAXCBkQCiIOTAEIggHkQOD0wAcIIB+EDggFkQQAiIDysZT6YAGEBZQsp5MAnCBwQBsIFiyJSxXBAMIgcFqvHEGDghxU+yAA)

PR #1253, opened 2025-03-28
Improve split type errors

den-churbanov:
This PR is similar to [this one](https://github.com/effector/effector/pull/1139) - it adds more explicit errors for split operator.

This PR solves next issues: #622, #769, #999 
I added some tests for match function inference, inline prepend calls and attach in default case.

Also i'm not sure about [narrower types edge cases](https://tsplay.dev/w2ykrW) - all highlighted cases must be an errors, but ts don't think so 🥲


zerobias (approved):
Amazing, thanks!

Issue #1252, opened 2025-03-27
Wrong getting from users array


smenshikov:
In the `filter` function we search by `id` field, but in the `fn` function we get by index from array. Looks like it is wrong and `fn` should be `fn: (users, id) => users.find((user) => user.id === id)!`
![Image](https://github.com/user-attachments/assets/056aaebc-821c-4c8c-a1c9-1588728af403)

PR #1263, opened 2025-05-06
docs: fix return value in function 

alekcena:
Issues #1252


alekcena:
set the correct issues

PR #1322, opened 2026-03-11
fix(sample): respect batch option for shape source

### LINKED ISSUE #1211: sample `source` option should respect `batch`

https://share.effector.dev/BIxXS9u2
https://stackblitz.com/edit/effector-vite-react-template-tbxm3h?file=src%2Fmodel.ts

There is a case for `batch: false` when store being consistently updated by high frequent events; sample use combine under the hood for `source: {$store}` case without passing `batch` to it, which leads to stale store value reads 

No stale value reads

anchmelev:
### Conventions
- [x] Please check your messages [against the guidelines](https://cbea.ms/git-commit/)

Closes #1211
Supersedes #1321

## Problem
`sample` has two source paths:
- `source: $store`
- `source: { list: $store }` (shape, internally uses combine)

With `batch: false`, shape-source path could read stale value in the same tick because internal combine sampler-read was still batched.

## Solution
- In `sample`, for shape-source:
  - keep `combine(source)` for `batch: true`
  - use `combineSource(shape, batch)` for `batch: false`
- In `combine`, pass `batch` to internal sampler read: `read(rawShape, true, batch)`.
- Follow-up types cleanup in second commit: replaced newly added `any` with `unknown` in this path.

## Result
No stale reads for shape-source with `batch: false`.

## Regression test
Added test for reported scenario:
- source shape + `batch: false`
- sequential same-tick updates
- expected state: `['hello', 'dudes']`

## Local checks
- `yarn test`
- `yarn build`

### DIFF
--- a/src/effector/__tests__/sample/sample.test.ts
+++ b/src/effector/__tests__/sample/sample.test.ts
@@ -854,3 +854,41 @@ describe('greedy deprecation', () => {
     )
   })
 })
+
+test('source shape respects batch option', () => {
+  const appStarted = createEvent()
+  const $list = createStore<string[]>([])
+
+  const initOne = createEvent()
+  const initTwo = createEvent()
+  const firstName = createEvent<string>()
+  const secondName = createEvent<string>()
+
+  sample({
+    clock: initOne,
+    fn: () => 'hello',
+    target: firstName,
+  })
+  sample({
+    clock: initTwo,
+    fn: () => 'dudes',
+    target: secondName,
+  })
+
+  sample({
+    clock: appStarted,
+    target: [initOne, initTwo],
+  })
+
+  sample({
+    clock: [firstName, secondName],
+    source: {list: $list},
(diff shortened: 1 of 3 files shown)


zerobias (approved):

zerobias:
Good work 👍 I personally can’t make release (but it’s looks like a patch release), so help from @sergeysova @AlexandrHoroshih or somebody from core team would be appreciated. Don’t forget to mention @anchmelev in release notes 😃

anchmelev:
Hi @sergeysova, @AlexandrHoroshih — gentle follow-up on this PR.

The fix addresses the stale read issue reported in #1211 for shape sources with `batch: false`, and the PR already includes a regression test for the reported case.

It was approved by @zerobias, all checks are passing, and he noted this looks like a patch-level change. Would appreciate a quick look at merge/release readiness when you have time. Thank you.

